PR #5, opened 2016-06-30
Client support for retrying requests

ekampf:
Client can retry requests in case of transport errors (usefull for avoiding timeout errors and other http stuff)

Also:
* added ability to specify default timeout for `RequestsHTTPTransport` and allow overriding the default timeout when calling `execute`.

* Added ability to have `RequestsHTTPTransport` post JSON requests

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -12,7 +12,7 @@ class RequestsHTTPTransport(HTTPTransport):
         """
         :param url: The GraphQL URL
         :param auth: Auth tuple or callable to enable Basic/Digest/Custom HTTP Auth
-        :param use_json: Choose whether to send the request body as a json or as form-urlencoded (Default: form-urlencoded)
+        :param use_json: Send request body as JSON instead of form-urlencoded
         :param timeout: Specifies a default timeout for requests (Default: None)
         """
         super(RequestsHTTPTransport, self).__init__(url, **kwargs)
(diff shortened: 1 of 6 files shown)


gql/transport/requests.py
```diff
@@ -8,21 +8,33 @@
...
+        :param use_json: Send request body as JSON instead of form-urlencoded
+        :param timeout: Specifies a default timeout for requests (Default: None)
+        """
         super(RequestsHTTPTransport, self).__init__(url, **kwargs)
         self.auth = auth
+        self.default_timeout = timeout
+        self.use_json = use_json
 
-    def execute(self, document, variable_values=None):
+    def execute(self, document, variable_values=None, timeout=None):
```
syrusakbary: I think is useful to maintain the `execute` method to be uniform across all transports/clients.
So the way of executing will be always the same regarding the transport used.

For that reason I think is better to move out the timeout from here and just use `self.default_timeout`.
Thoughts?

ekampf: @syrusakbary I have this use case:
I create a Client with default timeout of 360. It works in most cases but there's this 1 big call I'm making that takes a long time. I want to give that call 10 minutes timeout.

So specifically when executing that call I want to override the default timeout...  creating a separate Client for it is unnecessary...

`Client` already defines execute as `def execute(self, document, *args, **kwargs):` and forwards all the `kwargs` to the transport...  So I don't see why transports should not support custom kwargs

ekampf:
@syrusakbary  ?

syrusakbary:
I don't want to block the merge until I have a better solution.

Right now your proposal makes sense, I think there should be a better way for doing it, but until then... merging is the answer!

PR #10, opened 2016-10-22
JSONfy variables if not string

anibalsolon:
Gives the possibility to specify variables as a python dict.

One of the reasons to have variables in GraphQL is to avoid value injections like `{"id":"%s"} % (10)` directly on the query. So, to keep this feature, I've implemented this little PR.

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import requests
+import json
 from graphql.execution import ExecutionResult
 from graphql.language.printer import print_ast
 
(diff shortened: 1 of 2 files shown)


gql/transport/requests.py
```diff
@@ -24,9 +25,13 @@ def execute(self, document, variable_values=None, timeout=None):
         query_str = print_ast(document)
         payload = {
             'query': query_str,
-            'variables': variable_values or {}
         }
 
+        if type(variable_values) is str:
```
alexharveyeu: `if isinstance(variable_values, str):`

koliyo:
This would be really helpful!

PR #15, opened 2017-05-02
Added iso8601 to dev requirements to fix build.

afharvey:


alexharveyeu:
Travis is red because of some password being wrong.

I think this solves the problem on master: `ImportError: No module named 'iso8601'`.

artofhuman:
@Cito i think it may be closed since master is green.

Cito:
Thanks for reporting.

Looks like iso8601 must not be required explicitly any more. I also recently removed the dev_requirements file since they are already in the setup.py and I wanted to avoid duplication.

PR #16, opened 2017-05-12
set use_json default true in RequestsHTTPTransport

stegben:
Since we usually set `content-type`: `application/json` when we use graphql ?


ahussain3:
+1

Cito:
I think this makes sense. Quoting the docs: A standard GraphQL POST request should use the application/json content type, and include a JSON-encoded body.

PR #17, opened 2017-05-31
Added json dump if passing in a dict, added operation_name

cjmaio:
Checks if `query` (previously `query_str`) is a dict, and converts using `json.dumps` otherwise leaves it be.

Adds in the ability to pass in an `operation_name` to `execute`

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+import json
 import requests
 from graphql.execution import ExecutionResult
 from graphql.language.printer import print_ast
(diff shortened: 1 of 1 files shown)


gql/transport/requests.py
```diff
@@ -20,12 +21,14 @@ def __init__(self, url, auth=None, use_json=False, timeout=None, **kwargs):
...
-        query_str = print_ast(document)
+    def execute(self, document, variable_values=None, operation_name=None, timeout=None):
+        query = print_ast(document)
         payload = {
-            'query': query_str,
+            'query': json.dumps(query) if isinstance(query, dict) else query,
             'variables': variable_values or {}
         }
+        if operation_name:
+            payload.operationName = operation_name
```
barbieri: you may always pass `operationName`, using the same trick as done for variables:

```
 'operationName': operation_name or ''
```

xin7c:
I am using gql v0.1.0
But I could not find
 "if operation_name:
            payload.operationName = operation_name"
in source code.
So, I can't add operation_name like this:
    r = client.execute(query, variable_values=json.dumps(params), operation_name="query_from_me")

It take me An error:
TypeError: execute() got an unexpected keyword argument 'operation_name'

So ...  why ?

Cito:
Looks like this is outdated now, so I'm closing this. (Also, print_ast should always return str, so I don't understand the purpose of this. Please re-submit if you think it still makes sense.)

Cito:
The `operation_name` was good though, it has now been added in the v2 and v3 (master) branch. Should have been done earlier.

PR #18, opened 2017-06-23
Feature/fix build

ekampf:


Cito:
Looks like this is outdated and has been fixed long ago already, so closing this.

PR #24, opened 2018-01-06
Fix/Feature: Objects via params using DSL

itolosa:
Now it's possible to use dsl to build this type of queries:
```python
ds.Mutation.login.args(
    email='user@example.com',
    password={'digest': 'ac67a676', 'algorithm': 'sha-256'}
  ).select(ds.LoginResponse.status)
```


Cito:
Unfortunately, this has become so seriously outdated that it cannot be merged any more and I need to close this.

KingDarBoja:
@Cito to be honest with you, several of these PR changes were used on the linked PRs in order to fix the DSL feature.

PR #26, opened 2018-01-18
Expose the verify option of the requests library

robrkerr:
Modified from #19, which hasn't had recent activity. The only difference if here the default value for this new argument should give the same behaviour as before (`verify=True`).


KingDarBoja:
This can be closed @Cito

Cito:
Closing this, because it is already part of #63.

PR #33, opened 2018-04-30
Add support for custom scalar types

ahussain3:
In the GraphQL schema, it is possible to specify (CustomScalarTypes)[https://www.apollographql.com/docs/graphql-tools/scalars.html].

In order to specify a Custom Scalar Type, you need to define an object which implements the following interface:
* `serialize()`
* `parse_literal()`
* `parse_value()`
These methods allow GraphQL to translate between the serialized and deserialized versions of the custom scalar object.

However, when the GQL schema is passed around it includes only the names of the custom scalar types, and not the serialization/deserialization methods. After this PR is merged, you should be able to define your own classes on the client, and by giving them a `parse_value()` method you can use them as custom scalar types.

To configure the gql client to use your custom scalar types, you should do:
```
from gql import gql, Client
custom_types = {
    'DateTime': MyDateTimeClass
}

client = Client(schema=schema, transport=transport, custom_types=custom_types)
response = client.execute(query)
```

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -4,6 +4,7 @@ from graphql import parse, introspection_query, build_ast_schema, build_client_s
 from graphql.validation import validate
 
 from .transport.local_schema import LocalSchemaTransport
+from .response_parser import ResponseParser
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


Cito:
It's a bit sad that gql has been dormant and PRs like this did not receive feedback. I will try to get this merged into v3 (modern version), and if requested backport from there to v2 (legacy version). Also, I will probably make this a bit more general to support all kinds of leaf types, i.e. also Enum types.

PR #39, opened 2018-09-04
Added support for Python 3.6 and 3.7

clintonb:
Dropped support for Python 3.3, which reached end of life in September 2017


Cito:
This has already been added.

PR #45, opened 2019-02-19
Should pass kwargs into RequestsHTTPTransport requests method

cw1427:
We should allow the **kwargs to be passed into the RequestHTTPTransport by the init kwargs.  For example, we would like the requests to mark the SSL verify to be false.  So it could be easily to be passed. like :  

self.transport = RequestsHTTPTransport(url=self.url, auth=HTTPBasicAuth(self.account, self.pwd), use_json=True,
                                               verify=False)

### DIFF
--- a/gql/transport/http.py
+++ b/gql/transport/http.py
@@ -1,6 +1,6 @@
 class HTTPTransport(object):
 
-    def __init__(self, url, headers=None, cookies=None):
+    def __init__(self, url, headers=None, cookies=None, **kwargs):
         self.url = url
         self.headers = headers
         self.cookies = cookies
--- a/gql/transport/http.py
+++ b/gql/transport/http.py
@@ -4,3 +4,4 @@ class HTTPTransport(object):
         self.url = url
         self.headers = headers
         self.cookies = cookies
+        self.kwargs = kwargs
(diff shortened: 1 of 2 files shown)


scjody:
@syrusakbary Can this be merged? The CI failures appear unrelated to the code being added.

If there's something missing I'm happy to help.

ankeshanand:
It would be great if this is merged! One use case we have is passing `proxies` into the requests POST call. The current API prevents that.

ankeshanand:
@syrusakbary Any chance this could merged? It seems very self-contained and would enable a lot of use-cases.

chriddyp:
We'd love to see this merged. Let us know if there is anything we can do to help. cc'ing the latest committers @Cito @artofhuman

artofhuman:
I think we need to post patch version before merging new PRs, but now we dont have access to pypy. Maybe we need to release this lib under the new name on pypy. WDYT @Cito ?

Cito:
Yes, we're currently stuck because we can't push to PyPI and Syrus needs to hand over maintainer rights. He did this already with most other libraries in graphql-python, but gql was overlooked because at that time nobody seemed to care. Give us some time to contact Syrus and sort this out, he is always very busy. Creating a new name should be the last resort only.

PR #49, opened 2019-08-09
Feature/allow errors to be raised or returned

alexmclarty:
PR to allow:

- errors to be raised with a custom exception (`ResultError`) which will allow better handling by consumers of the `Client` class.
- errors to be returned along with data which will allow consumers of the `Client` class to act on all errors rather than the first in a list.

I'd like to get some feedback on the approach before merging.

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -7,6 +7,7 @@ from .transport.local_schema import LocalSchemaTransport
 
 log = logging.getLogger(__name__)
 
+
 class RetryError(Exception):
     """Custom exception thrown when retry logic fails"""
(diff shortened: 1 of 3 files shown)


leszekhanusz:
I though a lot about this proposal and now think that adding a raise_error flag in the client may not be our best option.

I made a new PR #96 with a proposed alternative.

- The TransportQueryError exception now returns all errors in an `errors` attribute.
- private functions `_execute` and `_subscribe` now exist in the sessions and can be used by the user if he wants to have results as ExecutionResult instances with errors instead of data and exceptions.

PR #56, opened 2020-02-08
Add anaconda cloud version badge

KingDarBoja:
The `graphql-python/gql` package is already updated to latest version available at pypi and anaconda cloud under the **conda-forge** channel. Would be great to showcase the badge so conda users can find it quickly.


Cito:
Thank you.

PR #60, opened 2020-02-23
Provide contributing docs, test suite fixes and return graphql errors

### LINKED ISSUE #40: Allow 400 status code if result contain valid graphql error message
The graphql server I am using return a HTTP 400 response for invalid requests. So I get both a 400 response, but with proper graphql error data in the body.

Using `graphiql` or similar I get the graphql error presented, but with `gql.client` I only get

```
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: ...
```

due to

```
request.raise_for_status()
```

in https://github.com/graphql-python/gql/blob/master/gql/transport/requests.py#L39

These errors should be returned to the user, not just throw a 400-exception

KingDarBoja:
- Place dev and test packages inside setup python file.
- Provide contributing docs for pip and conda developers.
- Add Makefile
- Add several tests in order to increase code coverage
- Fixes #40. **This was required in order to let both _test_no_schema_ and  _test_execute_result_error_ pass**.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,49 @@
+# Contributing
+
+Thanks for helping to make gql awesome!
+
+We welcome all kinds of contributions:
+
+- Bug fixes
+- Documentation improvements
+- New features
+- Refactoring & tidying
+
+
+## Getting started
+
+If you have a specific contribution in mind, be sure to check the [issues](https://github.com/graphql-python/gql/issues) and [pull requests](https://github.com/graphql-python/gql/pulls) in progress - someone could already be working on something similar and you can help out.
+
+
+## Project setup
+
+After cloning this repo, ensure dependencies are installed by running:
+
+```sh
+make dev-setup
+```
+
+## Running tests
+
+After developing, the full test suite can be evaluated by running:
+
+```sh
+make tests
+```
+
+## Development on Conda
+
+In order to run `tox` command on conda, you must create a new env (e.g. `gql-dev`) with the following command:
+
+```sh
+conda create -n gql-dev python=3.8
+```
+
+Then activate the environment with `conda activate gql-dev` and install [tox-conda](https://github.com/tox-dev/tox-conda):
+
+```sh
+conda install -c conda-forge tox-conda
+```
+
+Uncomment the `requires = tox-conda` line on `tox.ini` file and that's it! Run `tox` and you will see all the environments being created and all passing tests. :rocket:
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,38 +12,86 @@ We welcome all kinds of contributions:
 
 ## Getting started
 
-If you have a specific contribution in mind, be sure to check the [issues](https://github.com/graphql-python/gql/issues) and [pull requests](https://github.com/graphql-python/gql/pulls) in progress - someone could already be working on something similar and you can help out.
-
+If you have a specific contribution in mind, be sure to check the 
+[issues](https://github.com/graphql-python/gql/issues) 
+and [pull requests](https://github.com/graphql-python/gql/pulls) 
+in progress - someone could already be working on something similar 
+and you can help out.
 
 ## Project setup
 
-After cloning this repo, ensure dependencies are installed by running:
+### Development with virtualenv (recommended)
+
+After cloning this repo, create a virtualenv:
+
+```console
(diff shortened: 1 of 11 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+
+If you have a specific contribution in mind, be sure to check the [issues](https://github.com/graphql-python/gql/issues) and [pull requests](https://github.com/graphql-python/gql/pulls) in progress - someone could already be working on something similar and you can help out.
+
+
+## Project setup
+
+After cloning this repo, ensure dependencies are installed by running:
+
+```sh
+make dev-setup
```
Cito: Can't we just put the actual command `pip install -e .[test]` here? Hiding this step in a makefile makes intransparent what is happening in this step. You also don't want to frustrate people using Windows which does not have a make command. Also, you should recommend using a virtual env. Otherwise the dependencies will be all installed into the global one.

tests/starwars/test_dsl.py
```diff
@@ -84,67 +98,6 @@ def test_fetch_luke_query(ds):
     assert query == str(query_dsl)
 
 
-# def test_fetch_some_id_query():
```
Cito: I guess the idea was to implement these tests with gql. So I would leave them there as a reminder.

gql/transport/requests.py
```diff
@@ -36,7 +36,7 @@ def execute(self, document, variable_values=None, timeout=None):
             data_key: payload
         }
         request = requests.post(self.url, **post_args)
-        request.raise_for_status()
+        # request.raise_for_status()
```
Cito: I think this should be moved to the assertion below. At least in the case when there is no GraphQL result (the assertion fails), raise an error with `request.raise_for_status()` and then an unspecific "raise HTTPError('no GraphQ result')" to catch the case when there is no GraphQL result but the status is good.

Also, shouldn't it be `response = ` instead of  `request = `?

Cito:
Thank you @KingDarBoja -  I had a quick glance and commented some issues. Note that the idea is to eventually merge gql-next in to gql and release it as the next major version.

tests/starwars/test_dsl.py
KingDarBoja: Those tests already exist on the `test_query.py` file, which make use of gql instead of plain string, that's why I removed them.

gql/transport/requests.py
KingDarBoja: Yeah, wasn't sure about moving it around due to the assertion.

CONTRIBUTING.md
KingDarBoja: I followed the `graphene-django` documentation as base for gql docs, including the Makefile but you're right, it is better to provide the actual command instead of using a alias on the Makefile.

For Windows users, I found it is better to use either the conda development steps or we should rely on some `Make.bat` as equivalent.

Cito: Right, but not all people on Windows are using Conda. It should also not be emphasized too much because otherwise the Windows user might think they *must* use Conda to develop gql.

tests/starwars/test_dsl.py
Cito: Ah, ok then.

CONTRIBUTING.md
KingDarBoja: I will make the clarifications at the night so users can have multiple choices to setup a development environment.

gql/transport/requests.py
Cito: What do you think about something like this?

```python
response = requests.post(self.url, **post_args)
try:
    result = request.json()
    if not isinstance(result, dict):
        raise ValueError
except ValueError:
    result = {}
if 'errors' not in result and 'data' not in result:
    request.raise_for_status()
    raise HTTPError("Expected a GraphQL result, but got: {}".format(
        response.content), response=response)
return ExecutionResult(
    errors=result.get('errors'), data=result.get('data'))
```

I think that is much better than using an assertion.

Cito: Or maybe just

    raise HTTPError("Server did not return a GraphQL result", response=response)

to avoid extremely long error messages (e.g. when you get an HTML page instead of GraphQL), since the response will be an attribute of the exception and response.content can be picked up from there.

KingDarBoja: Could work, as long as it return an error response for the user which does not interfer with GraphQL errors.

PR #61, opened 2020-02-28
[WIP] Tests/coverage tests

KingDarBoja:
Aiming to improve test coverage and also provide little improvements to the `gql` package.

- [x] References #24: Now the DSL support using mutation as stated on the referenced PR.
- [X] Add dsl mutation test.
- [x] Add http transport raise for status error test.

### DIFF
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -4,8 +4,8 @@ from functools import partial
 import six
 from graphql.language import ast
 from graphql.language.printer import print_ast
-from graphql.type import (GraphQLField, GraphQLList,
-                          GraphQLNonNull, GraphQLEnumType)
+from graphql.type import (GraphQLField, GraphQLList, GraphQLNonNull, GraphQLEnumType)
+from graphql.utils.ast_from_value import ast_from_value
 
 from .utils import to_camel_case
 
@@ -31,7 +31,7 @@ class DSLSchema(object):
         return self.execute(query(*args, **kwargs))
 
     def mutate(self, *args, **kwargs):
-        return self.query(*args, operation='mutate', **kwargs)
+        return self.query(*args, operation='mutation', **kwargs)
 
     def execute(self, document):
         return self.client.execute(document)
@@ -126,10 +126,12 @@ def field(field, **args):
     raise Exception('Received incompatible query field: "{}".'.format(field))
 
 
-def query(*fields):
+def query(*fields, **kwargs):
+    if 'operation' not in kwargs:
+        kwargs['operation'] = 'query'
     return ast.Document(
         definitions=[ast.OperationDefinition(
-            operation='query',
+            operation=kwargs['operation'],
             selection_set=ast.SelectionSet(
                 selections=list(selections(*fields))
             )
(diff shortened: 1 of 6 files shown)


KingDarBoja:
@Cito This is ready for review :D

gql/dsl.py
```diff
@@ -117,29 +102,32 @@ def __str__(self):
         return print_ast(self.ast_field)
 
 
-def field(field, **args):
+def selection_field(field, **args):
     if isinstance(field, GraphQLField):
+        # TODO: Pass name argument
```
Cito: Can you add a test that passes a GraphQLField to select(), to get that case covered?

tests/starwars/test_dsl.py
```diff
@@ -122,3 +122,56 @@ def test_hero_name_query_result(ds):
         }
     }
     assert result == expected
+
+
+# TODO: Keep the function name or switch to schema query names?
```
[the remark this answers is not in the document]
KingDarBoja: As most names on `test_dsl.py` (I would rather say the same for `test_query.py`) make use of the query name for the test function, it is harder to guess which one is testing a specific functionality.

In this case, I wasn't sure about following the same convention and went with specifying the function that is being tested.

I do feel like it is a better approach but added the `TODO` notice to get some feedback about the naming convention for the test suite.

gql/dsl.py
KingDarBoja: No problem!

tests/starwars/test_dsl.py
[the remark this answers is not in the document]
Cito: Ah, ok. Keep it like that, I like how you named the test.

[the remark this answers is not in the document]
Cito: Btw, since you're at it already: The `test_invalid_field_on_type_query` has a bad indentaton (the assert should not be in the `raises` block and bad spelling "doesnt".

gql/dsl.py
KingDarBoja: The init function at DSLField requires it:
```python
    def __init__(self, name, field):
        self.field = field
        self.ast_field = ast.Field(name=ast.Name(value=name), arguments=[])
        self.selection_set = None
```
I tried this query:
```
def test_pass_grapqhl_field_on_query(ds):
    result = ds.query(
        ds.Query.hero.select(
            queryType.fields.get('hero')
        )
    )
    expected = {
        'hero': {
            'name': 'R2-D2'
        }
    }
    assert result == expected
```

But no success :/ @Cito

tests/starwars/test_dsl.py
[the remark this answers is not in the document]
KingDarBoja: Done! I will push the changes after adding the last dsl test.

gql/dsl.py
Cito: Right, a bare GraphQLField has no name. Not sure how this was intended to be used. Maybe we should simply remove that whole if branch and the `**args`.

Regarding the missing coverage in requests. The "raise Value" happens e.g. when the server returned JSON that is not an object, but an array. The "requests.HTTPError" happens when the server does not report an valid GraphQL result, but with a status of OK.

KingDarBoja: Sounds good! I will provide last two missing test on another PR.

Cito: Ok. Is this PR ready to be merged then from your side?

KingDarBoja: Yeah! Ready for merge!

PR #32, opened 2018-04-27
Update Readme to include more example code

ahussain3:
As discussed in [https://github.com/graphql-python/gql/issues/29#issuecomment-384472839], it would be great if the readme included more example code to help people get started.

PR #63, opened 2020-03-06
Refactor RequestsHttpTransport Class

### LINKED ISSUE #14: Skip https certificate verification
Hi,

It seems it's not possible to skip https certificate verification without hardcode in the lib.

Skip works fine if to add in:
class RequestsHTTPTransport() -> execute: post_args -> add 'verify': False

And it seem's it would be perfect to add https verification as  additional method's param.

### LINKED ISSUE #29: Any example code?

### LINKED ISSUE #43: Should pass kwargs into RequestsHTTPTransport requests method
Looks like the RequestsHTTPTransport did not pass the kwargs into the requests. It should be flexible to pass the kwargs into the requests when __init__ the transport or execute the execute function.

  post_args = {
            'headers': self.headers,
            'auth': self.auth,
            'timeout': timeout or self.default_timeout,
            data_key: payload
        }
        request = requests.post(self.url, **post_args)
        request.raise_for_status()

For example:  post_args.update(self.kwargs)

### LINKED ISSUE #54: No documentation
Tried replicating the two tests against a publicly available end-point, but they fail. Also tried lifting implementation ideas from Apollo documentation into Python, but the differences were not intuitively discoverable. Unable to proceed with any implementation.

KingDarBoja:
This PR provides some minor changes to the `RequestsHttpTransport` class.

- [x] Remove the `HttpTransport` class and pass directly the `headers`, `url` and `cookies` keyword arguments to the `RequestsHttpTransport` class.
- [x] Pass `verify` kwarg to the `RequestsHttpTransport` class. `True` by default.
**Fixes #45, Fixes #19, Fixes #43, Fixes #14.**
- [x] Provide more documentation examples.
**Fixes #29, Fixes #54 based on the PR #32, Fixes #32**
- [x] Provide test example of reading `.graphql` schema files.
- [x] Create `Transport` abstract class and provide test for `NotImplementedError` if not properly instantiated.

I had to reference both issues and PR in order to clear the repository from redundant or stale stuff.


KingDarBoja:
@Cito this is ready for review!

gql/transport/requests.py
```diff
@@ -4,21 +4,31 @@
 from graphql.execution import ExecutionResult
 from graphql.language.printer import print_ast
 
-from .http import HTTPTransport
 
-
-class RequestsHTTPTransport(HTTPTransport):
-    def __init__(self, url, auth=None, use_json=False, timeout=None, **kwargs):
+class RequestsHTTPTransport(object):
```
Cito: Since you bypassed HTTPTransport and deleted that file I think it makes sense to rename this class HTTPTransport now for simplicity.

But maybe you can add an abstract base class "Transport" with the execute method for both LocalSchemaTranport and HTTPTransport. Later it can be also used for type checking the transport argument of the Client.

Also, docstrings for these classes would be nice to have. You can mention that HTTPTransport uses requests and the POST method for instance.

tests/fixtures/graphql/sample.graphql
```diff
@@ -0,0 +1,15 @@
+type User {
+    id: ID!
+    username: String
+    first_name: String
```
Cito: Better use camelCase here?

tests/starwars/test_query.py
```diff
@@ -343,12 +343,12 @@ def test_check_type_of_luke(client):
 
 def test_parse_error(client):
     result = None
-    with pytest.raises(Exception) as excinfo:
+    with pytest.raises(Exception) as excInfo:
```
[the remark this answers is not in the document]
KingDarBoja: You're right, I took a quick look at pytest docs and found the variable being written as `exec_info` instead of `excInfo`. Not sure why I used the latter convention.

gql/transport/requests.py
KingDarBoja: That's a good idea! The only thing I don't understand at this point is the existance of **LocalSchemaTransport** and **RequestsHttpTransport** as I have no idea what's the main difference between them.

Also, by looking at some features from #24, I do see extra transport classes were suggested (and added) like as **BatchHttpTransport**, **SessionHttpTransport** and so on.

A quick search on Google leads me to think those are _extra transport layers for GraphQL_ but not sure about this.

Cheers!

tests/starwars/test_query.py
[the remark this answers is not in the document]
Cito: It's just PEP8 convention.

gql/transport/requests.py
Cito: The LocalschemaTransport uses a local, executable GraphQL schema instead of querying a remote GraphQL server.

Cito: Extra transport classes should then all inherit from the abstract base Transport class.

Cito: In a future version (when we require Python 3.8+), maybe we can use a [protocol](https://www.python.org/dev/peps/pep-0544/) instead of an abstract base class.

tests/fixtures/graphql/sample.graphql
KingDarBoja: Done!

gql/transport/requests.py
KingDarBoja: I have made the changes, at the end, created the `Transport` file which stores the abstract class. Also added tons of docstrings and tested locally with sphinx for hyperlinks and such stuff.

sydoluciani:
@Cito any update ? this pull request fixes couple of long standing issues. 

@syrusakbary  can you please add me to contributors so I can can contribute to graphql-python libraries.

By the way, the contribution link is broken:
https://github.com/graphql-python/gql/blob/master/contributing.md

KingDarBoja:
@sydoluciani Don't get me wrong but it is a bad idea to add someone to the maintainer team without previous contributions to the library. You can contribute by opening pull requests in order to fix or provide new features to the library.

Also, Syrus is no loner active on the graphql-python Team as he is doing other stuff so please don't ping him.

gql/transport/requests.py
Cito: Big thanks and kudos for working on this!

gql/transport/requests.py
```diff
@@ -1,26 +1,66 @@
...
+        url,  # type: str
+        headers=None,  # type: Dict[str, Any]
+        cookies=None,  # type: Union[Dict[str, Any], RequestsCookieJar]
+        auth=None,  # type: AuthBase
+        use_json=False,  # type: bool
+        timeout=None,  # type: int
+        verify=True,  # type: bool
+        **kwargs  # type: Any
+    ):
+        """Create a HTTPTransport using requests library to execute GraphQL queries on remote servers.
```
Cito: Suggestion:

class docstring:

    """Transport to execute GraphQL queries on remote servers. 

    The transport uses the requests library to send HTTP POST request.
    """

init docstring:

    """Initialize the transport with the given request parameters.

    :prarm url: ... etc
    """

gql/transport/requests.py
```diff
@@ -1,26 +1,66 @@
...
         self.auth = auth
-        self.default_timeout = timeout
         self.use_json = use_json
+        self.default_timeout = timeout
+        self.verify = verify
+        self.kwargs = kwargs
 
     def execute(self, document, variable_values=None, timeout=None):
+        # type: (Node, dict, int) -> ExecutionResult
+        """Execute the provided document AST for the provided remote server.
```
Cito: "Execute the provided docuemtn AST against the configured remote server"

gql/transport/requests.py
```diff
@@ -1,26 +1,66 @@
...
+        This make use of requests library within the POST method to perform http calls to the remote server.
```
Cito: A bit simpler: "This uses the requests library to perform a HTTP POST request to the remote server."

tests/fixtures/graphql/sample.graphql
```diff
@@ -0,0 +1,15 @@
...
+    fullName: String
+}
+
+type Query {
+    user(id: ID!): User
+}
+
+schema {
+    query: Query
+}
```
Cito: looks like a newline is missing here in the last line

tests/test_client.py
```diff
@@ -1,8 +1,33 @@
...
+    }
+    ''')
+
+
+def test_request_transport_not_implemented(http_transport_query):
+    class RandomTransport(Transport):
+        def execute(self):
+            super(RandomTransport, self).execute(http_transport_query)
+
+    with pytest.raises(NotImplementedError) as err:
```
Cito: `exc_info` instad of `err` for consistency

gql/transport/transport.py
```diff
@@ -0,0 +1,20 @@
+import abc
```
Cito: Maybe put this `transport/__init__.py` instead of creating a new file, so you can 'import Transport from gql.transport' or just 'import Transport from .' in local_schema and requests instead of 'import Transport from gql.transport.transport' which looks a bit ugly. And the submodules need the base class anyway.

Cito:
@sydoluciani The contribution link is actually one of the things that has been fixed in this PR.

PR #65, opened 2020-03-09
Docs subscriptions

KingDarBoja:
Based on [this graphene PR](https://github.com/graphql-python/graphene/pull/1107/files) and reading about [GraphQL sample schema](https://github.com/apollographql/starwars-server/blob/master/data/swapiSchema.js), I tried to replicate the subscription feature using gql library.

There were big issues as this isn't well documented but this PR should provide insight for end users to use GraphQL subscriptions along with gql.

The only issue is that **asyncio, async/await features are not supported on Python 2.7.** 
Also, the **yield** keyword will throw a `SyntaxError` on Python 3.5.

Also, the `Episode` enumerable was using integers instead of strings values, so I replaced them to follow the sample GraphQL schema.

For last, this PR also provides support for passing `GraphQLInputObjectType` and `GraphQLInputObjectField` if using DSL based on #24.

Cheers!

### DIFF
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -1,10 +1,15 @@
 from functools import partial
 
 import six
-from graphql import GraphQLInputObjectType, GraphQLInputObjectField
 from graphql.language import ast
 from graphql.language.printer import print_ast
-from graphql.type import (GraphQLEnumType, GraphQLList, GraphQLNonNull)
+from graphql.type import (
+    GraphQLEnumType,
+    GraphQLInputObjectField,
+    GraphQLInputObjectType,
+    GraphQLList,
+    GraphQLNonNull,
+)
 from graphql.utils.ast_from_value import ast_from_value
 
 from .utils import to_camel_case
@@ -31,7 +36,7 @@ class DSLSchema(object):
         return self.execute(query(*args, **kwargs))
 
     def mutate(self, *args, **kwargs):
-        return self.query(*args, operation='mutation', **kwargs)
+        return self.query(*args, operation="mutation", **kwargs)
 
     def execute(self, document):
         return self.client.execute(document)
@@ -54,7 +59,9 @@ class DSLType(object):
         if camel_cased_name in self.type.fields:
             return camel_cased_name, self.type.fields[camel_cased_name]
 
-        raise KeyError('Field {} does not exist in type {}.'.format(name, self.type.name))
+        raise KeyError(
+            "Field {} does not exist in type {}.".format(name, self.type.name)
+        )
 
 
 def selections(*fields):
@@ -63,7 +70,6 @@ def selections(*fields):
 
 
 class DSLField(object):
-
     def __init__(self, name, field):
         self.field = field
         self.ast_field = ast.Field(name=ast.Name(value=name), arguments=[])
@@ -88,10 +94,7 @@ class DSLField(object):
             arg_type_serializer = get_arg_serializer(arg.type)
             serialized_value = arg_type_serializer(value)
             self.ast_field.arguments.append(
-                ast.Argument(
-                    name=ast.Name(value=name),
-                    value=serialized_value
-                )
+                ast.Argument(name=ast.Name(value=name), value=serialized_value)
             )
         return self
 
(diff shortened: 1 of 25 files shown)


KingDarBoja:
@Cito Can you provide some feedback in this one? Thank you 😃

tests/starwars/schema.py
```diff
@@ -1,24 +1,25 @@
...
-from .fixtures import getCharacters, getDroid, getFriends, getHero, getHuman, updateHumanAlive
+from .fixtures import createReview, getCharacters, getDroid, getFriends, getHero, getHuman, reviewAdded
 
 episodeEnum = GraphQLEnumType(
     'Episode',
     description='One of the films in the Star Wars Trilogy',
     values={
         'NEWHOPE': GraphQLEnumValue(
-            4,
+            value='NEWHOPE',
```
Cito: Is there a reason why you changed the values? When they are different, the tests can detect better when the wrong thing is returned (name instead of value) is returned. Btw, these numeric values are also used by the GraphQL.js and graphq-core tests.

KingDarBoja: > Based on this graphene PR and reading about GraphQL sample schema, I tried to replicate the subscription feature using gql library.

I thought they used string values for the enumerable, as seen [here](https://github.com/apollographql/starwars-server/blob/master/data/swapiSchema.js#L361).

Cito: They construct a schema from SDL, so they can't define values. In that case, the values are the names. But since we have a schema with server side values, I think we should also use them. The schema in GraphQL.js that uses the numeric values is [here](https://github.com/graphql/graphql-js/blob/master/src/__tests__/starWarsSchema.js).

KingDarBoja: Okay, I will bring back the old values instead of using string as currently does.

gql/dsl.py
```diff
@@ -1,6 +1,7 @@
 from functools import partial
 
 import six
+from graphql import GraphQLInputObjectType, GraphQLInputObjectField
```
Cito: You should merge this with the other import from graphql.type

tests/starwars/fixtures.py
```diff
@@ -93,8 +112,8 @@ def getFriends(character):
 
 def getHero(episode):
     if episode == 5:
-        return luke
-    return artoo
+        return humanData.get('1000')
```
Cito: Since humaData is not changed (any more), you can simply return "luke" again.

KingDarBoja:
Looks like mypy got mad about some definitions!

Cito:
I merge this anyway and will fix the mypy issues then.

Cito:
@KingDarBoja: Maybe you can make your next PRs and commits a bit more focused, i.e. better create more small PRs and commits than one large one, and keep the janitor work, cleanups etc. separate from the code changes and bug fixes.

Cito:
The mypy issues are fixed now in 2aeffa50e41b186eaba2d4cfaa0c65d15061bdb4. I simply silenced most warnings, some were caused by incomplete type hints in graphql-core 2. But we should not waste too much time with this. The typing stuff is much better in graphql-core 3 where we can also use real Python 3 type hints.

KingDarBoja:
> @KingDarBoja: Maybe you can make your next PRs and commits a bit more focused, i.e. better create more small PRs and commits than one large one, and keep the janitor work, cleanups etc. separate from the code changes and bug fixes.

Thanks for the suggestion, I will keep that on mind. Small PRs to address a specific bug fix or feature while cleanup and all those stuff on another PR.

> The mypy issues are fixed now in [2aeffa5](https://github.com/graphql-python/gql/commit/2aeffa50e41b186eaba2d4cfaa0c65d15061bdb4). I simply silenced most warnings, some were caused by incomplete type hints in graphql-core 2. But we should not waste too much time with this. The typing stuff is much better in graphql-core 3 where we can also use real Python 3 type hints.

I was going to ask at the Slack channel about spending some extra time (at least for me) to go around graphql-core-legacy and fix some issues to clean the repo from staled issues and pull requests due to development of graphql-core-3 being the current focus.

Shall I do it or just ignore any related issue and instead wait for the gql-next migration plan?

Cito:
Some people are waiting for issues and PR in graphql-core-legacy to be processed. If you can help with that, this will be good.

PR #67, opened 2020-03-19
Publish wheels to pypi

borgstrom:
Hi 👋 

This PR adds support to build & publish universal wheels (`.whl`) to pypi so that users do not need to "build from source" when installing.

See: https://pythonwheels.com/

Thanks!

cc @Cito


Cito:
Thank you!

Issue #62, opened 2020-03-04
Roadmap getting us from legacy to modern version


KingDarBoja:
Python 2 has reached EOL and no further development will be done as stated on the [official Python release schedule page for 2.7](https://www.python.org/dev/peps/pep-0373/)

> Being the last of the 2.x series, 2.7 will receive bugfix support until 2020. Support officially stops January 1 2020, but the final release will occur after that date.
> 
> Planned future release dates:
> 
> 2.7.18 code freeze January, 2020
> 2.7.18 release candidate early April, 2020
> 2.7.18 mid-April, 2020

Also, Python 3.5 support will stop at 2020-09-13 as seen at [status-of-python-branches](https://devguide.python.org/#status-of-python-branches), so would be great to drop support for Python 3.5 in order to push forward and use 3.6+ features on the code (like f-strings).

Cito:
This should happen together with the switch from GraphQL-Core version 2 to version 3.

@ekampf has been already working on that in graphql-python/gql-next. That work should be merged back here and released as new major version of gql. As long as we support the GraphQL-Core legacy version 2, we should also support the legacy Python 2.

My suggestion is to fast forward the version number of gql to version 2, and then release the version based on GraphQL-Core 3 as version 3. This versioning scheme has also been used with other GraphQL-Core based libraries such as Graphene.

KingDarBoja:
Sounds good, the only issue (which I have no idea) is the merge step, as I see there is not too many shared stuff between each version, perhaps the upcoming gql version (or v3) is made from scratch, right?

And regarding the fast forward to v2 following the semantic versioning process, is there any schedule date?

Cito:
Regarding the merge step, give me some time to look into this and decide how to best merge the two since now gql has moved forward. With "from scratch" you still mean based on the code and API of v2, or do you want to create a completely new API?

Regarding the timing for v2, we should first triage the open issues and PRs and check if there are bugs that should be fixed or features that should be added to v2, or be postponed or rejected for good reasons. It would be great if you could help with that.

KingDarBoja:
Regarding the "from sratch", I was actually asking about how it was done since it doesn't seem to be the same repository organization (files, directories, functions) on the quick look between `gql-next` and `gql`.

I am open to keep contributing to the already existing open issues, which some of them **have been addressed** on some PRs, but I do feel like those can be done in a better way.

Once again, not the most skilled on the GraphQL spec but willing to improve and keep contributing 👍

[...]

PR #70, opened 2020-03-25
Async usage with new websockets and http async transports using asyncio

leszekhanusz:
Necessitate Python 3.6 and asyncio

### New Client
* can be used as before synchronously
```python
client.execute(query)
```
* but can also be used asynchronously
```python
async with Client(...) as session:
    await session.execute(query)
```
* supports existing sync http transport (requests) (only synchronously)
* supports also new async transports (websockets and aiohttp)
* Supports subscriptions with the websockets transport
* synchronously with an infinite python generator
```python
for result in client.subscribe(query):
    ... result is a dict here
```

* OR asynchronously with an async generator.
```python
async with Client(...) as session:
    async for result in session.subscribe(query):
        ... result is a dict here
```
* for async transports, fetching schema from transport is done when executing the first query

### New Websockets async transport

Allows a connection to an Apollo GraphQL websocket endpoint using the [apollo subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md) protocol

Require the great [websockets library](https://websockets.readthedocs.io/en/stable/intro.html)

Supports:

* queries, mutations AND subscriptions
* SSL encryption
* multiple queries using the same connection
* multiple subscription in parallel on the same connection using asyncio tasks

### New HTTP async transport

Supports thanks to the [aiohttp library](https://docs.aiohttp.org/en/stable/):
* http headers
* http cookies
* Basic auth
* SSL

### gql-cli

* Added gql-cli script to execute graphql queries from the command line to an http or websocket endpoint.


yiblet:
thanks for doing this work, I was looking exactly for these features!

KingDarBoja:
Been busy due to real-life stuff but will look into this PR as soon as possible.

setup.py
```diff
@@ -6,7 +6,8 @@
     'six>=1.10.0',
     'graphql-core>=2,<3',
     'promise>=2.0,<3',
-    'requests>=2.12,<3'
+    'requests>=2.12,<3',
+    'websockets>=8.1,<9'
```
KingDarBoja, line 10: From pypi, websockets support Python 3.6.1 and higher. This one should be appended to the `install_requires` list. See line 20 for an example with `pytest-asyncio`.

leszekhanusz: Done. Now tests are passing. Will need to add tests to raise the coverage now.

leszekhanusz: Also now that I understand asyncio a little better I think it is best to use Futures intead of Queues for the listeners. Will need to change that also so don't accept this PR just yet.

KingDarBoja: No problem mate :)

KingDarBoja:
This is looking great, of course requires several reviews from the graphql-python team but glad to see this effort 🚀

setup.py
leszekhanusz: In fact queues are really needed for slow consumers ...

maeglindeveloper:
Hi everyone, is there any example for that new feature :) !?

leszekhanusz:
> Hi everyone, is there any example for that new feature :) !?
> I see also that the merging is blocked, do you have an ETA of release ?
> Thanks for the good work ! +1

There are examples added in the README.md
You can see them here [https://github.com/leszekhanusz/gql](https://github.com/leszekhanusz/gql) in the "Websockets transport with asyncio" section

leszekhanusz:
Note: should I rebase my pull request and squash the small fix commits ?

This would avoid the merge commit but would rewrite history in this PR

maeglindeveloper:
@leszekhanusz, the best way is to have it review no ?  @syrusakbary, wdyt ?
Then new release can be done ?

maeglindeveloper:
Any updates on this ? :

KingDarBoja:
@leszekhanusz don't worry about that, at the end this PR will be squashed into a single commit to keep things clean. I will review this PR after discussing with **graphql-python** folks about the next release (v3). 

Please read [Cito comment](https://github.com/graphql-python/gql/issues/62#issuecomment-619544513) about what I am talking about.

maeglindeveloper:
Keep us update :) ! I'm looking forward to use that new feature.
Great job guys

sneko:
Awesome @leszekhanusz! That's cool this library will benefit from this, congrats!

Just curious, may I missed something but will we be able to deal with the initial payload at init?

Thank you,

leszekhanusz:
> Just curious, may I missed something but will we be able to deal with the initial payload at init?
Do you mean the payload we are sending in the connection_init message ? Currently this payload is empty but we could add a parameter init_payload to the transport for it.

sneko:
Yeah I was thinking about this.

In another project using Javascript I had to use it to pass some init parameters but also passing the access token to allow or not establishing the connection. At this time I had no other choice for the access token (not even using a header I remember).

sneko:
Woow, such reactivity! Thank you for your work. I hope it will pass the review to try all of this 👍 🎉

Cito:
Thanks again for this contribution. After creating a v2.x branch for the legacy support, and declaring the master to be the modern v3 branch, I merged this into master, and after fixing a small merge problem, tests are green. I will do some clean-up and modernization work in the master branch now.

PR #73, opened 2020-04-24
Retry logic change proposal

fclairamb:
First of all, thank you for this great library that we use in many of our tools, libraries and services at @habx.

The `Client` code should let the `Transport` handle the retry logic.

This would simply improve the library because:
- It allows clients to customize the created `requests.Session` behavior to implement their own retry logic. We currently are wrapping your code in ours simply to add some exponential backoff (which is pretty standard).
- This makes the calling code shorter 
- We currently end-up performing retries on transports like the `LocalSchemaTransport` where it doesn't make sense

I think we should also remove the `retries` parameter in the `Client.execute` code but that would mean breaking the current API. It all depends on how you feel about that.

I'm open to any suggestions to improve this PR and help us deal with this issue.

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -25,7 +25,7 @@ class Client(object):
         type_def=None,
         transport=None,
         fetch_schema_from_transport=False,
-        retries=0,
+        retries=0,  # We should remove this parameter and let the transport level handle it
     ):
         assert not (
             type_def and introspection
--- a/gql/client.py
+++ b/gql/client.py
@@ -93,3 +93,13 @@ class Client(object):
                 retries_count += 1
 
         raise RetryError(retries_count, last_exception)
+
+    def close(self):
+        """Close the client and it's underlying transport"""
+        self.transport.close()
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
--- a/gql/transport/__init__.py
+++ b/gql/transport/__init__.py
@@ -24,7 +24,8 @@ class Transport:
     def close(self):
         """Close the transport
 
-        This method doesn't have to be implemented unless the transport would benefit from it,
-        for example any long-lived TCP Connection like HTTP Keep-Alive or Websocket.
+        This method doesn't have to be implemented unless the transport would benefit from it.
+        This is currently used by the RequestsHTTPTransport transport to close the session's
+        connection pool.
         """
         pass
--- a/gql/transport/__init__.py
+++ b/gql/transport/__init__.py
@@ -20,3 +20,17 @@ class Transport:
         raise NotImplementedError(
             "Any Transport subclass must implement execute method"
         )
+
+    def close(self):
+        """Close the transport
+
+        This method doesn't have to be implemented unless the transport would benefit from it,
+        for example any long-lived TCP Connection like HTTP Keep-Alive or Websocket.
+        """
+        pass
+
+    def __enter__(self):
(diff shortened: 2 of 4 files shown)


KingDarBoja:
I don't see any problem with the idea as the current `retries` parameter would be deprecated to avoid breaking changes while transitioning into `gql-v3`.

I will take a look at it, seems like `pytest-asyncio` got a new release therefore Travis CI job crashed.

KingDarBoja:
@fclairamb mate, can you rebase your branch with latest changes from master? That should fix the travis CI job.

fclairamb:
Liting issue is fixed, there's just a little problem that is causing me some pains.

A test fail on python 3.5:
```
_______________________ test_http_transport_verify_error _______________________
[...]
        with pytest.warns(Warning) as record:
            client.execute(http_transport_query)
>       assert len(record) == 1
E       assert 2 == 1
E        +  where 2 = len(WarningsChecker(record=True))
```

It looks like it's a python 3.5 specific issue but it's not, it happens on any version. You can reproduce it in the latest version of python as-well.

The underlying error is something like that:
```
ResourceWarning("unclosed <ssl.SSLSocket fd=13, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('192.168.168.142', 64269), raddr=('34.83.64.96', 443)>")
```

The issue comes from this:
- Each HTTPTransport now has a `requests.Session` object
- This `requests.Session` keeps a pool of connections
- When we run multiple tests we end-up with the previous tests leaving a "leak" of `SSLSocket` still waiting to be closed.

What I ended up doing is adding the ability to `close()` the client that will itself close its underlying `Transport`. The `close()` method has a default empty implementation on the `Transport`. And as such adding some `__enter__` and `__exit__` method to be able to use through as context manager.

In real-life scenarios, we most probably never need to `close()` the `Transport` or `Client`.

There are alternatives:
- Closing the session after each query (that would be strange because it would prevent us from doing any kind of `Keep-Alive`)
- Using `requests.Session` only if we want to. It's easy to do but makes the code a bit confusing IMO
- Allowing to pass a `requests.Retry` instance that can be passed with each method but that kinds of cancels the idea of having generic session that is both standard and easy to extend.

I feel bad complexifying the code-base, still I think it makes sense to handle it like that.

fclairamb:
Hi @KingDarBoja,

So I fixed all the linting, potential connection leak, and coverage issues.

 What do you think ?

KingDarBoja:
@fclairamb Thanks for the contribution, checking it right now as it seems a good change that doesn't involve a breaking change.

fclairamb:
Great! Thank you!

dmsehuang:
Hi @fclairamb , I encountered `unclosed connection` issue while unit testing my client application.

Sample error code:
```
ResourceWarning: unclosed <ssl.SSLSocket fd=748, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('x.x.x.x', 50754), raddr=('x.x.x.x', 443)>
```

source code:

```python
_transport = RequestsHTTPTransport(url, headers)
client = Client(transport=_transport, fetch_schema_from_transport=True)
query = gql('''some valid GQL query''')
client.execute(query)
```

I tried both methods to close the connection:
```python
transport.close()
client.transport.close()
```
but none of them works.

Could you provide some guidance here?

PR #74, opened 2020-04-25
chore: bump pytest and pytest-asyncio dependencies

KingDarBoja:
This will fix the Travis CI job as `pytest-asyncio` got a new update which relies on the `pytest` v5.4.0+ API.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,14 +11,14 @@ install_requires = [
 
 tests_require = [
     'coveralls==1.11.1',
-    'pytest==4.6.9',
+    'pytest==5.4.1',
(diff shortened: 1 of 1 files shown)


Cito:
This is still using pytest 4.6 because it is the last one that supports Python 2, and the current gql v0.x which will be relased as v2 still supports Python 2 and needs to be tested with that. We must really move forward with #62.

KingDarBoja:
@Cito In this case, I just bumped the version for Python 3.6 and above whereas keep the old version for Python 2.7 / 3.5 so those can still be supported, nothing was changed besides that.

PR #75, opened 2020-04-28
Documentation update

fclairamb:
Thank you for merging my previous PR.

It seems like you should push users to use this new API.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,3 +35,4 @@ deploy:
   on:
     tags: true
     python: 3.8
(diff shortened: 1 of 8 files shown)


KingDarBoja:
Hello @fclairamb 

I would suggest putting a deprecation notice regarding the `retries` argument on the client, something like:

> The `retries` parameter on the Client class is deprecated, you should pass it at the Transport class to avoid unnecessary retries call on `LocalSchemaTransport` (as example).

Cheers!

PR #76, opened 2020-04-29
Allows for custom/any HTTP method (other than POST)

MonsieurV:
Hi,

I wanted to use this library for a project, but I can't run the queries against the GraphQL API that I use (Prismic GraphQL).

The reason: it only support `GET` HTTP method (and return 404 for `POST`) ([see Prismic doc](https://prismic.io/docs/graphql/getting-started/intro-to-the-graphql-api#14_2-a-graphql-api-that-uses-get)).

As the method was hardcoded as `POST` in the library, I was unable to make it work passing a `method` kwargs to`RequestsHTTPTransport` :

```python
client = gql.Client(
        transport=gql.transport.requests.RequestsHTTPTransport(
            url="https://<repo>.prismic.io/graphql",
            headers={
                "Content-type": "application/json",
                "Authorization": "Token <token>",
                "Prismic-ref": "<ref>",
            },
            method="GET",
        ),
        fetch_schema_from_transport=True,
    )
```

Here a PR for making that possible. By default, the HTTP method remains `POST`; but for those that need/want, they can pass another method to `RequestsHTTPTransport`.

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -29,6 +29,7 @@ class RequestsHTTPTransport(Transport):
         timeout=None,  # type: int
         verify=True,  # type: bool
         retries=0,  # type: int
+        method="POST",  # type: str
         **kwargs  # type: Any
     ):
         """Initialize the transport with the given request parameters.
@@ -43,6 +44,7 @@ class RequestsHTTPTransport(Transport):
             the server's TLS certificate, or a string, in which case it must be a path
             to a CA bundle to use. (Default: True).
         :param retries: Pre-setup of the requests' Session for performing retries
+        :param method: HTTP method used for requests. (Default: POST).
         :param kwargs: Optional arguments that ``request`` takes. These can be seen at the :requests_: source code
             or the official :docs_:
 
@@ -56,6 +58,7 @@ class RequestsHTTPTransport(Transport):
         self.use_json = use_json
         self.default_timeout = timeout
         self.verify = verify
+        self.method = method
         self.kwargs = kwargs
 
         # Creating a session that can later be re-use to configure custom mechanisms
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -48,6 +48,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *.cover
+gql-dev/
```
KingDarBoja, line 51: Why this one is needed?

gql/transport/requests.py
```diff
@@ -56,6 +58,7 @@ def __init__(
         self.use_json = use_json
         self.default_timeout = timeout
         self.verify = verify
+        self.method = method
```
KingDarBoja, line 61: Should be good to put a little note on the readme about this. I think I will update the readme after merging this.

KingDarBoja:
For anyone looking at the **requests** package documentation regarding the function arguments: [requests.Session.request](https://requests.readthedocs.io/en/master/api/#requests.Session.request).

.gitignore
MonsieurV: This is indeed not directly related to the issue at hand, but was required (to not have a clumsy `git status` output) for me when using the `virtualenv` as recommended in [CONTRIBUTING](https://github.com/graphql-python/gql/blob/master/CONTRIBUTING.md#development-with-virtualenv-recommended):

```sh
virtualenv gql-dev
```

(Create a `gql-dev` directory for the virtuelenv)

gql/transport/requests.py
MonsieurV: Yep, maybe I let you do it, because I don't see a clear place for documentation arguments. (I could add it to the third example with a comment?)

.gitignore
KingDarBoja: That shouldn't be ignored as I haven't seen anyone with the same issue and also, the virtualenv name can vary between contributors.

Can you remove this please?

MonsieurV: Sure I can.

Please note that any following the guidelines and using the same virtualenv command will have the same issue... except if the `CONTRIBUTING` is updated to use the more common `venv` virtuelenv name (which is already in the `.gitignore`).

Cito: Changing the CONTRIBUTING.md to use `venv` or `.venv` (which seems to be even more common nowadays and is in `.gitignore` as well) is a good idea.

PR #78, opened 2020-05-06
Doc update + deprecation notice

fclairamb:
This is a follow-up on #75, because I messed-up the previous branch.

This includes:
- A README update
- A deprecation warning so that users progressively switch to retries on the transport level (if applicable)


fclairamb:
I fixed the linting issue.

Issue #92, opened 2020-05-25
crash when connecting gql to hasura


mariusnita:
Hi, I'm trying this client with hasura and it crashes when executing the introspection query due to 
https://github.com/graphql-python/gql/blob/master/gql/transport/aiohttp.py#L109

If that line is just `operation_name` instead of `operation_name or ""` then it works. Hasura breaks when the operationName is set to `""`.

Not positive if this is a hasura issue or a gql issue.

leszekhanusz:
Hi,

Could you please post here the complete error message ?

The PR #94 should fix your issue

mariusnita:
Thanks!

The relevant part of the stack trace was this:
```
  File ".../lib/python3.8/site-packages/gql/client.py", line 88, in execute_async
    return await session.execute(document, *args, **kwargs)
  File ".../lib/python3.8/site-packages/gql/client.py", line 279, in execute
    await self.fetch_and_validate(document)
  File ".../lib/python3.8/site-packages/gql/client.py", line 247, in fetch_and_validate
    await self.fetch_schema()
  File ".../lib/python3.8/site-packages/gql/client.py", line 298, in fetch_schema
    self.client.schema = build_client_schema(self.client.introspection)
  File ".../lib/python3.8/site-packages/graphql/utilities/build_client_schema.py", line 56, in build_client_schema
    raise TypeError(
TypeError: Invalid or incomplete introspection result. Ensure that you are passing the 'data' attribute of an introspection response and no 'errors' were returned alongside: None.
```

And this was the hasura reply
```py
{'errors': [{'extensions': {'path': '$', 'code': 'validation-failed'}, 'message': 'no such operation found in the document: ""'}]}
```

leszekhanusz:
Fixed in version v3.0.0a1

PR #94, opened 2020-05-25
aiohttp transport - optional variables and operation name

leszekhanusz:
- works the same as the requests transport
- fix connecting to servers which do not support setting operationName to "" (#92)


leszekhanusz:
> By the way, this is more like a refactor than a feature as the behaviour is almost the same but to avoid any unexpected server crash as requested at #92.

Indeed. I removed "Feature" from the title

PR #95, opened 2020-05-26
Use less generic exceptions

leszekhanusz:


scripts/gql-cli
```diff
@@ -27,7 +27,7 @@ async def main():
     elif scheme in ["http", "https"]:
         transport = AIOHTTPTransport(url=args.server)
     else:
-        raise Exception("URL protocol should be one of: http, https, ws, wss")
+        raise ValueError("URL protocol should be one of: http, https, ws, wss")
```
stabacco, line 30: Wondering if an [UnknownProtocol](https://docs.python.org/3.8/library/http.client.html?highlight=unknownprotocol#http.client.UnknownProtocol) exception would be a better fit in this case.

leszekhanusz: I think ValueError is fine.

If we compare with other libraries:
[aiohttp raise a ValueError](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/web_urldispatcher.py#L740)
[yarl raises a ValueError](https://github.com/aio-libs/yarl/blob/master/yarl/__init__.py#L361)
[requests raises a custom InvalidSchema which inherits ValueError](https://github.com/psf/requests/blob/master/requests/exceptions.py#L76)

PR #96, opened 2020-05-30
Return all errors in TransportQueryError exception

leszekhanusz:
* Add an errors attribute in the TransportQueryError to get all the errors in the exception
* expose private _execute and _subscribe functions to get ExecutionResult data directly if needed

### DIFF
--- a/gql/transport/exceptions.py
+++ b/gql/transport/exceptions.py
@@ -1,3 +1,6 @@
+from typing import Any, List, Optional
+
+
 class TransportError(Exception):
     pass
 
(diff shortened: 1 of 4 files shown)


leszekhanusz:
Note: instead of private `_execute` and `_subscribe` functions, we could have public functions with a more explicit name (something like `execute_execution_result` and `subscribe_execution_result` but it is a little too much verbose for my taste)

gql/transport/exceptions.py
```diff
@@ -22,9 +25,15 @@ class TransportQueryError(Exception):
     This exception should not close the transport connection.
     """
 
-    def __init__(self, msg, query_id=None):
+    def __init__(
+        self,
+        msg: str,
+        query_id: Optional[int] = None,
+        errors: Optional[List[Any]] = None,
```
KingDarBoja, line 32: Everything is okay, but I think we could provide a better type as `TransportQueryError` is only used at `client.py` where it receives `Optional[List[GraphQLError]]` and `websockets.py` which receives `Any` as type (due to json parsing).

Not sure if possible.

leszekhanusz: The problem is that currently there is no validation done of the errors returned from the different possible types of backends. We could try to convert the errors received to `GraphQLError` but it could break existing code if the error received is not compatible.

KingDarBoja: Gotcha! Let's then merge this as it is.

PR #98, opened 2020-06-02
Add gitpod config

Raptus369:
this commit adds support for Gitpod.io, a free automated
dev environment that makes contributing and generally working on GitHub
projects much easier. It allows anyone to start a ready-to-code dev
environment for any branch, issue and pull request with a single click.

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,10 @@
+FROM gitpod/workspace-mysql
+                    
+USER gitpod
+
+# Install custom tools, runtime, etc. using apt-get
+# For example, the command below would install "bastet" - a command line tetris clone:
+#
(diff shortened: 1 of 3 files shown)


KingDarBoja:
As non-user of Gitpod, I did a quick search on the [gitpod.io](http://gitpod.io/) website and noticed the free pricing plan which is restricted to 50 hours per month, so does that means any dev has this kind of restriction while working on this cloud-based (I think it is) dev tool?

leszekhanusz:
Isn't it spam ? There is nothing gql related in this PR. Those files could have been added to any repo. User was created yesterday ...

Cito:
Agree with you both. Only if this would be a completely free offer (not limited time) this *might* be something we would support with a badge.

PR #104, opened 2020-06-14
Feat: Add support for custom scalar types [updated]

KingDarBoja:
Updated #33 with latest changes from master. Had to open my own branch as I don't have permission to push to source repo.


leszekhanusz:
Also why do we need to provide a class with a `parse_value` method ? Isn't it simpler to just provide a method directly ?

KingDarBoja:
> Also why do we need to provide a class with a `parse_value` method ? Isn't it simpler to just provide a method directly ?

Not sure what was the reasoning behind this but probably it was done in order to keep most of the boilerplate code on its own class so any extra stuff can be added later to extend the type adapter functionality.

leszekhanusz:
Some thoughts:

* why do we limit ourselves to convert only scalars ? Could it be more useful to be also able to convert GraphQLObjects ?
* As of now, what we have will convert scalars using only the answer from the server and the schema and does not take into account the query. This could cause problems:
    * if the same name is used as a root field for query, mutation and subscription
    * This can't work if the user is using aliases

KingDarBoja:
> why do we limit ourselves to convert only scalars ? Could it be more useful to be also able to convert GraphQLObjects ?

Indeed, this should support any kind of leaf types like as Enums or Objects, this PR wanted to bring the old one with the latest features of gql so I didn't bother to add the rest of types to keep the review simple.

> As of now, what we have will convert scalars using only the answer from the server and the schema and does not take into account the query.

Never though about that, maybe we should perform the conversion using the AST and checking each node kind to find the correct and perform the operation on it.

leszekhanusz:
The proposition of @KingDarBoja to use the Visitor pattern on the document AST has been implemented in PR #256.
So I am going to close this PR as obsolete.

PR #107, opened 2020-06-28
100% tests coverage!

leszekhanusz:
dsl: Fix select used multiple times on same field
local_schema: Fix correct return of errors in subscriptions

### DIFF
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -84,7 +84,7 @@ class DSLField(object):
         added_selections = selections(*fields)
         if selection_set:
             selection_set.selections = FrozenList(
-                selection_set.selections + [selection for selection in added_selections]
+                selection_set.selections + list(added_selections)
             )
         else:
             self.ast_field.selection_set = SelectionSetNode(
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -84,7 +84,7 @@ class DSLField(object):
         added_selections = selections(*fields)
         if selection_set:
             selection_set.selections = FrozenList(
-                selection_set.selections + added_selections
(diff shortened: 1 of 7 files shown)


gql/transport/local_schema.py
```diff
@@ -50,27 +50,16 @@ async def execute(
...
 
         The results are sent as an ExecutionResult object
         """
 
-        subscribe_result = subscribe(self.schema, document, *args, **kwargs)
+        subscribe_result = await subscribe(self.schema, document, *args, **kwargs)
 
         if isinstance(subscribe_result, ExecutionResult):
-            yield ExecutionResult
+            yield subscribe_result
```
KingDarBoja, line 61: How was that working before? Anyway, good catch!

leszekhanusz: Yes, that was a typo but it happened only in case of errors with subscriptions using the local_schema transport. So it was not really used apparently.
That shows the importance of test coverage!

gql/dsl.py
```diff
@@ -84,7 +84,7 @@ def select(self, *fields):
         added_selections = selections(*fields)
         if selection_set:
             selection_set.selections = FrozenList(
-                selection_set.selections + added_selections
+                selection_set.selections + [selection for selection in added_selections]
```
Cito, line 87: Good catch!

A list comprehension is not necessary, though. You can do
```py
                selection_set.selections + list(added_selections)
```
or maybe
```py
                chain(selection_set.selections, added_selections)
```
with
```py
from itertools import chain
```

leszekhanusz: You are right. I went with your first solution as it is more readable.

Issue #108, opened 2020-06-30
Connection ack error when keep alive message arrives first


gregsanz182:
Hi!. Currently I'm using this package to test a Hasura project, but when the WebsocketsTransport waits for the ack message on initial connection, the Hasura server first responds with a keep alive message. Because of this, the WebsocketsTransport raises a TransportProtocolError exception, given that it waits for a `connection_ack`, not a `ka` message.

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/sanchez/Projects/graphql_client.py", line 829, in test_subscription
    asyncio.run(self.something())
  File "/usr/lib/python3.8/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
    return future.result()
  File "/home/sanchez/Projects/graphql_client.py", line 839, in something
    async with Client(
  File "/home/sanchez/Projects/venv/lib/python3.8/site-packages/gql/client.py", line 161, in __aenter__
    await self.transport.connect()
  File "/home/sanchez/Projects/venv/lib/python3.8/site-packages/gql/transport/websockets.py", line 508, in connect
    raise e
  File "/home/sanchez/Projects/venv/lib/python3.8/site-packages/gql/transport/websockets.py", line 503, in connect
    await self._send_init_message_and_wait_ack()
  File "/home/sanchez/Projects/venv/lib/python3.8/site-packages/gql/transport/websockets.py", line 193, in _send_init_message_and_wait_ack
    raise TransportProtocolError(
gql.transport.exceptions.TransportProtocolError: Websocket server did not return a connection ack
```

Here you can see the order from the messages sent by my Hasura server. In this case, I'm testing using the Hasura web console:

![image](https://user-images.githubusercontent.com/9099404/86171647-0dcf6d00-baeb-11ea-996a-1fabe09ba656.png)

As you can see, the `ka` message comes first, so TransportProtocolError rejects this message, and raises the above exception. Maybe TransportProtocolError needs to ignore `ka` messages when wating for ack message

leszekhanusz:
Thank you for your bug report.
Could you please check that the PR #110 solves your problem ?

hugbubby:
> Thank you for your bug report.
> Could you please check that the PR #110 solves your problem ?

How do you replace your local version of gql in a python venv with the version from this pull request?

KingDarBoja:
[...]

PR #110, opened 2020-07-04
Ignore keepalives before connection_ack in websockets transport

leszekhanusz:
This should fix bug #108

### DIFF
--- a/gql/transport/websockets.py
+++ b/gql/transport/websockets.py
@@ -175,6 +175,23 @@ class WebsocketsTransport(AsyncTransport):
 
         return answer
 
+    async def _wait_ack(self) -> None:
+        """Wait for the connection_ack message. Keep alive messages are ignored
+        """
+
+        while True:
+            init_answer = await self._receive()
+
+            answer_type, answer_id, execution_result = self._parse_answer(init_answer)
+
+            if answer_type == "connection_ack":
+                return
+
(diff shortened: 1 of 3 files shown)


hugbubby:
plz merg

gql/transport/websockets.py
```diff
@@ -175,6 +175,23 @@ async def _receive(self) -> str:
...
+
+            answer_type, answer_id, execution_result = self._parse_answer(init_answer)
+
+            if answer_type == "connection_ack":
+                return
+
+            if answer_type != "ka":
+                raise TransportProtocolError(
+                    "Websocket server did not return a connection ack"
+                )
```
KingDarBoja, line 193: Shouldn't be `"Webscocket server did not return a keep-alive"` ?

leszekhanusz: No, we are waiting here for a connection_ack (just above)
If we don't receive a connection_ack, we check if it was instead a keep-alive.
If it is not a keep-alive here, it means that it is not a keep-alive and not a connection_ack so we raise the exception same as before.
If it is a keep-alive, then we continue in the loop and will listen to the next message received hoping to receive a connection_ack next.

KingDarBoja: Fair enough, got it wrong while reviewing it.

PR #113, opened 2020-07-12
type_def argument of Client is now deprecated.

leszekhanusz:
The Client of gql allows us now to pass a local schema either:

- with the `schema` argument if the schema is a GraphQLSchema type
- with the `type_def` argument if the schema is passed as a string

But this is basically the same thing expressed with different types.

I propose to simplify the documentation that the `type_def` argument becomes deprecated and that the local schema is passed using the `schema` argument for both types.

To summarize, you should use `schema=` now instead of previously `type_def=` in the Client.

The schema argument will accept either:
- a string (previously passed to the type_def argument)
- a GraphQLSchema (like before)

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -1,6 +1,6 @@
 import asyncio
 import warnings
-from typing import Any, AsyncGenerator, Dict, Generator, Optional, Union, cast
+from typing import Any, AsyncGenerator, Dict, Generator, Optional, Union
 
 from graphql import (
     DocumentNode,
@@ -31,7 +31,7 @@ class Client:
     ):
         assert not (
             type_def and introspection
-        ), "Cannot provide introspection type definition at the same time."
+        ), "Cannot provide introspection and type definition at the same time."
 
         if type_def:
             assert (
@@ -51,7 +51,7 @@ class Client:
 
         if isinstance(schema, str):
             type_def_ast = parse(schema)
-            schema = cast(GraphQLSchema, build_ast_schema(type_def_ast))
+            schema = build_ast_schema(type_def_ast)
 
         if transport and fetch_schema_from_transport:
             assert (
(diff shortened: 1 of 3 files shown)


gql/client.py
```diff
@@ -31,22 +32,33 @@ def __init__(
         assert not (
             type_def and introspection
         ), "Cannot provide introspection type definition at the same time."
```
KingDarBoja, line 34: Not related to this PR but I think there is a missing connector `and` on the message.

Should be:
`Cannot provide introspection and type definition at the same time.`

gql/client.py
```diff
@@ -31,22 +32,33 @@ def __init__(
...
         if introspection:
             assert (
                 not schema
             ), "Cannot provide introspection and schema at the same time."
             schema = build_client_schema(introspection)
-        elif type_def:
+
+        if isinstance(schema, str):
+            type_def_ast = parse(schema)
+            schema = cast(GraphQLSchema, build_ast_schema(type_def_ast))
```
KingDarBoja, line 54: Why the cast? As far as I know, the [**build_ast_schema**](https://github.com/graphql-python/graphql-core/blob/3267c4c5c5ea06968beb59f16425de4a4b835ee1/src/graphql/utilities/build_ast_schema.py#L23) returns a `GraphQLSchema` type.

leszekhanusz: I had a mypy error earlier. But now it seems it is not necessary anymore. Well spotted.

gql/client.py
leszekhanusz: Indeed, I made the correction

leszekhanusz:
> I would rather have removed the `type_def` argument and mark as breaking change for gql v3.

We could leave it as deprecated now and remove it once we are at the stable 3.0 ?

KingDarBoja:
> > I would rather have removed the `type_def` argument and mark as breaking change for gql v3.
> 
> We could leave it as deprecated now and remove it once we are at the stable 3.0 ?

Well I am okay with that as long as it doesn't stay for more than 2 minor releases after stable 😄

PR #114, opened 2020-07-12
Doc update readme

leszekhanusz:
* Add warnings regarding 2.x version
* Better explanation on how to use a local schema
* Example of GraphQL query using variable values
* Very brief example of the DSL module
* Improved Async advanced usage example
* Added backoff module for retries in the advanced example
* Removed RequestsHTTPTransport example

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,9 @@ Plays nicely with `graphene`, `graphql-core`, `graphql-js` and any other GraphQL
 
 GQL architecture is inspired by `React-Relay` and `Apollo-Client`.
 
+> **WARNING**: Please note that the following documentation describes the current version which is currently only available as a pre-release
+> The documentation for the 2.x version compatible with python<3.6 is available in the [2.x branch](https://github.com/graphql-python/gql/tree/v2.x)
+
 [![travis][travis-image]][travis-url]
 [![pyversion][pyversion-image]][pyversion-url]
 [![pypi][pypi-image]][pypi-url]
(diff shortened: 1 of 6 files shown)


leszekhanusz:
Note for maintainers: the PR #112 and #113 should be merged before this one.

README.md
```diff
@@ -26,118 +29,187 @@ GQL architecture is inspired by `React-Relay` and `Apollo-Client`.
...
 ```python
-from gql import gql, Client
+from gql import gql, Client, AIOHTTPTransport
 
-from .someSchema import SampleSchema
+# Select your transport with a defined url endpoint
+transport = AIOHTTPTransport(url="https://countries.trevorblades.com/")
 
+# Create a GraphQL client using the defined transport
+client = Client(transport=transport, fetch_schema_from_transport=True)
```
KingDarBoja, line 47: As basic usage, would be great to keep minimal the example code by excluding the transport in this case.

A new section below this one should explain the transport option with an example as currently does and mention the other available transports.

README.md
```diff
@@ -26,118 +29,187 @@ GQL architecture is inspired by `React-Relay` and `Apollo-Client`.
...
+for result in client.subscribe(query):
+    print (result)
 ```
 
-If you have a local schema stored as a `schema.graphql` file, you can do:
+> **Note**: the websockets transport can also execute queries or mutations, it is not restricted to subscriptions
+
+### Execute on a local schema
+
+It is also possible to execute queries against a local schema (so without a transport).
```
KingDarBoja, line 162: Since I mentioned how the basic usage would be without the transport, maybe we should move this section after basic usage?

Perhaps mention that the basic usage uses a local transport and then show the example with transports?

README.md
leszekhanusz: I bet that more than 99% of users need the gql library to execute GraphQL queries from a server. Do you know one person which uses gql to execute a local schema ? That is the reason that I moved this functionality lower in the readme. I find it confusing if it is at first. IMHO the first example should be a very simple example of a query on an http transport like I did.

KingDarBoja: You got a point.

PR #116, opened 2020-07-18
DOC add note to warn about running basic example with asyncio

leszekhanusz:
DOC add note to warn about running basic example with an asyncio event loop running

Small fixes for the assertion messages

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ result = client.execute(query)
 print(result)
 ```
 
+> **WARNING**: Please note that this basic example won't work if you have an asyncio event loop running. In some
+> python environments (as with Jupyter which uses IPython) an asyncio event loop is created for you. In that case you
+> should use instead the example in the [AIOHTTPTransport](#HTTP-async-transport) section.
+
 ### Local schema validation
 
 It is possible to validate a query locally either using a provided schema or by using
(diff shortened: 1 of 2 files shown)


njbernstein:
FYI: asyncio example must be run in python 3.7.

njbernstein:
The code in the documentation still doesnt run for me but the following does:
```
from gql import gql, AIOHTTPTransport, Client
import asyncio

async def main():

    transport = AIOHTTPTransport(url='https://countries.trevorblades.com/graphql')

    # Using `async with` on the client will start a connection on the transport
    # and provide a `session` variable to execute queries on this connection
    async with Client(
        transport=transport,
        fetch_schema_from_transport=True,
        ) as session:

        # Execute single query
        query = gql('''
            query getContinents {
              continents {
                code
                name
              }
            }
        ''')

        result = await session.execute(query)
        print(result)

await main()from gql import gql, AIOHTTPTransport, Client
import asyncio

async def main():

    transport = AIOHTTPTransport(url='https://countries.trevorblades.com/graphql')

    # Using `async with` on the client will start a connection on the transport
    # and provide a `session` variable to execute queries on this connection
    async with Client(
        transport=transport,
        fetch_schema_from_transport=True,
        ) as session:

        # Execute single query
        query = gql('''
            query getContinents {
              continents {
                code
                name
              }
            }
        ''')

        result = await session.execute(query)
        print(result)

await main()
```

Issue #89, opened 2020-05-17
Create separate documentation using Sphinx


Cito:
We should create separate documentation using Sphinx, since the README is getting too large.

The documentation can then be made availabel online as GitHub pages deployed with an GitHub action, or via ReadTheDocs. RTD has the advantage that it can create versioned docs if we need that.

captaincoordinates:
Versioned docs would be a major benefit because "the following documentation describes the current version which is currently only available as a pre-rel[e]ase". I don't want to risk using the alpha version in my application and I wasted time trying to understand why I could not import from gql.transport.aiohttp (it didn't exist in v2)

PR #117, opened 2020-07-19
chore: generate docs with sphinx

KingDarBoja:
References #89 

So far so good, to generate the docs just run `sphinx-build -b html -EW docs docs/_build/html` and navigate to `docs/_build/index.html` and that's it!

However, the index is not listing anything whereas the modindex isn't being generated.


tox.ini
```diff
@@ -47,6 +47,12 @@ deps = -e.[dev]
 commands =
     mypy gql tests
 
+[testenv:docs]
+basepython = python3.8
+deps = -e.[dev]
+commands =
+    sphinx-build -b html -nEW docs docs/_build/html
```
leszekhanusz, line 54: If I run this command I get the error:
```Warning, treated as error:
/home/leszek/miniconda3/envs/gql-dev/lib/python3.8/site-packages/gql/client.py:docstring of gql.client.AsyncClientSession._execute::py:class référence cible non trouvée : graphql.language.ast.DocumentNode
```

But the documentation is correctly generated if I run `make html` in the docs folder

gql/transport/__init__.py
```diff
@@ -1,4 +1,15 @@
 from .async_transport import AsyncTransport
 from .transport import Transport
+from .aiohttp import AIOHTTPTransport
```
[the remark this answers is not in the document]
KingDarBoja: Ah, I was asking myself why those imports weren't at `__init__` file and I didn't saw the warning because I didn't ran the tests.

tox.ini
KingDarBoja: I already asked Cito on the Slack channel since I have no idea why those are generated. Seems like `graphql-core` has something similar and then nitpicking each one on the `conf.py` file. I tried the same but always got that error.

KingDarBoja:
> * is it possible to hide private methods in the docs ?  (starting with an underscore)

Yes, in this case would be manually importing every function/class per module instead of dumping all as I did. In fact, that was my original approach but as I wasn't sure how many stuff was going to be displayed on the docs, I went for the generic solution.

KingDarBoja:
If the structure is okay, then would be great to start moving the readme into a subsection called "usage" like [graphql-core docs](https://graphql-core-3.readthedocs.io/en/latest/usage/index.html) does.

tox.ini
KingDarBoja: @leszekhanusz run again but without the `n`:

`sphinx-build -b html -nEW docs docs/_build/html` Now these errors are gone!

leszekhanusz:
Note: I create the gql project on readthedocs.io and I think the webhook is in place so if I understand correctly, once we merge this PR it should be visible on gql.readthedocs.io

leszekhanusz:
First I still need to correct the references now...

leszekhanusz:
Note: I improved the class references but I removed the DSL module reference for now (no time and it will probably change...)

leszekhanusz (approved):

KingDarBoja:
Outstanding documentation, everything seems to be well documented. I will provide a separate section for DSL in the short term.

PR #127, opened 2020-08-02
Update the import for AIOHTTPTransport

Min-Chen:
Update the import for AIOHTTPTransport to reflect the code change.


leszekhanusz:
The transports can be imported directly from gql now with the latest version (with PR #112), there is no need to revert the documentation change.

KingDarBoja:
Marking as invalid as this change is not needed. Also closing this PR too to keep the repository clean, next time should be good to open a issue to validate if changes are needed :)

PR #131, opened 2020-08-12
Fixed the bug where a nested GraphQLInputObjectType causing infinite get_arg_serializer recursive calls

JBrVJxsc:
Sometimes a `GraphQLInputObjectType` could have a field that has/contains the same type of it, in short, it is nested.

As `get_arg_serializer` is essentially a DFS and it did not handle this potential infinite recursive case.

This PR addressed this by caching/returning the serializer that was already created instead of stepping into the rabbit hole.

In the unit test, I added a test case `test_nested_input_with_old_get_arg_serializer` to reproduce the error. Feel free to remove it if you think testing with the good case is sufficient enough.


JBrVJxsc:
Opened another PR that has cleaner commits.

leszekhanusz:
Don't worry about clean commits, all the commits in the PR are squashed anyway when it is merged.

PR #132, opened 2020-08-13
Fixed the bug where a nested GraphQLInputObjectType causing infinite get_arg_serializer recursive calls

JBrVJxsc:
Sometimes a `GraphQLInputObjectType` could have a field that has/contains the same type of it, in short, it is nested.

As `get_arg_serializer` is essentially a DFS and it did not handle this potential infinite recursive case.

This PR addressed this by caching/returning the serializer that was already created instead of stepping into the rabbit hole.

In the unit test, I added a test case `test_nested_input_with_old_get_arg_serializer` to reproduce the error. Feel free to remove it if you think testing with the good case is sufficient enough.

### DIFF
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -105,7 +105,7 @@ class DSLField(object):
             arg = self.field.args.get(name)
             if not arg:
                 raise KeyError(f"Argument {name} does not exist in {self.field}.")
-            arg_type_serializer = get_arg_serializer(arg.type)
+            arg_type_serializer = get_arg_serializer(arg.type, known_serializers=dict())
             serialized_value = arg_type_serializer(value)
             added_args.append(
                 ArgumentNode(name=NameNode(value=name), value=serialized_value)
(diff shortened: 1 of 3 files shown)


gql/dsl.py
```diff
@@ -151,17 +152,37 @@ def serialize_list(serializer, list_values):
...
+visited_arg_types: Dict[Any, SerializerWrapper] = {}
+
+
 def get_arg_serializer(arg_type):
     if isinstance(arg_type, GraphQLNonNull):
         return get_arg_serializer(arg_type.of_type)
     if isinstance(arg_type, GraphQLInputField):
         return get_arg_serializer(arg_type.type)
     if isinstance(arg_type, GraphQLInputObjectType):
+        if arg_type in visited_arg_types:
```
leszekhanusz, line 176: I admit that I don't really understand all the dsl code completely, but is the SerializerWrapper class really necessary ?
Isn't it equivalent to:

```python
    if isinstance(arg_type, GraphQLInputObjectType):                                                                    
        if arg_type in visited_arg_types:                                                                               
            return visited_arg_types[arg_type]                                                                          
                                                                                                                        
        visited_arg_types[arg_type] = None                                                                              
                                                                                                                        
        serializers = {k: get_arg_serializer(v) for k, v in arg_type.fields.items()}                                    
                                                                                                                        
        visited_arg_types[arg_type] = lambda value: ObjectValueNode(                                                    
            fields=FrozenList(                                                                                          
                ObjectFieldNode(name=NameNode(value=k), value=serializers[k](v))                                        
                for k, v in value.items()                                                                               
            )                                                                                                           
        )                                                                                                               
                                                                                                                        
        return visited_arg_types[arg_type]
```

gql/dsl.py
```diff
@@ -151,17 +152,37 @@ def serialize_list(serializer, list_values):
...
+
+    def set_serializer(self, serializer):
+        self._serializer = serializer
+        return self._serializer
+
+    def __call__(self, *args, **kwargs):
+        return self._serializer(*args, **kwargs)
+
+
+visited_arg_types: Dict[Any, SerializerWrapper] = {}
```
leszekhanusz, line 167: I don't like the idea to add a global variable here which is never cleaned. Maybe this could be an attribute of DSLSchema instead ?
This would need a refactor, adding a reference to DSLSchema to the DSLType and DSLField classes, and putting the method get_arg_serializer inside the DSLSchema class ?

gql/dsl.py
JBrVJxsc: The wrapper is necessary. It sets an anchor while there is no actual serializer's body finalized.
Your code wouldn't work as the infinite recursion happens at `serializers = {k: get_arg_serializer(v) for k, v in arg_type.fields.items()}`, it will never get a chance to hit `visited_arg_types[arg_type] = lambda value: ObjectValueNode(`, thus `if arg_type in visited_arg_types:` will always be `False`.

JBrVJxsc: Ignore my previous comment. I think you are right.

gql/dsl.py
JBrVJxsc: I've removed the global variable and made the `visited_arg_types` as a required argument with the name `known_serializers` into `get_arg_serializer`. So that we don't need to refactor that many places, but also allowed multiple `get_arg_serializer` invocation within the same context (using one dictionary passed in):
known_serializers = {}
serializer_1 = get_arg_serializer(type_1, known_serializers)
serializer_2 = get_arg_serializer(type_2, known_serializers)
serializer_3 = get_arg_serializer(type_3, known_serializers)

leszekhanusz: This is indeed much better. OK for me

Issue #130, opened 2020-08-11
Websockets transport stuck during the close step of a connection


akvadrako:
I am using the gql from master and experiencing an occasional deadlock. The library I'm using is calling `client.execute()` with a websocket transport. Sometimes I see an error like this: `Exception in subscribe: GeneratorExit()` and then `client.execute()` never returns.

`WebSocketTransport.execute()` does return an `ExecutionResult`, i.e. `first_result` is set. Also `AsyncClientSession.execute()` returns.

Maybe it's failing to close the connection?

leszekhanusz:
Which version of python are you using ?

akvadrako:
3.8.5

leszekhanusz:
Could you try to figure out where this is stuck ? Maybe in the `await self._wait_closed.wait()` last line of websockets.py ?

And if it is possible also give a log using

```python
import logging
logging.basicConfig(level=logging.INFO)
```

akvadrako:
I will try to find a more specific location and upload some logs. Unfortunately, it's difficult to reproduce.

akvadrako:
Turns out it didn't take long. Here are some logs (sorry for the mess):

https://gist.github.com/b031bc6aa97ca1a15d1d363e339e3fee

Also, indeed it does get stuck on that line in `wait_closed()`. I added some print statements in `_close_coro` and `_clean_close`. Normally I see `_close_coro_` run. When it fails, I don't see it.

leszekhanusz:
Thank you for the log, this is indeed quite strange.

Normally after we receive the `complete` message from the server, the connection should be closed cleanly by sending the `connection_terminate` message. Apparently we don't do that for the last request in the logs. The close coroutine might be stuck in the `await self._send_connection_terminate_message()` for some reason and because of that, we are stuck waiting for this coroutine to stop in the `wait_closed` method.

The connection seems to be left open as we continue to receive the keep-alive messages.

We probably need to add timeouts for some actions in the `_close_coro` and put some of these actions in try blocks to protect us in case of problems.

That being said for your use case you should really reuse the same connection instead of creating another connection for each request. This would solve the problem for you.
For that, you should use `async with client as session:` only once in the code and use `session.execute` for each of your requests. See the advanced usage in the README.

akvadrako:
Thanks. A timeout sounds fine - I already have the closing timeout set to 1.

[...]

PR #133, opened 2020-08-17
Fix race condition in websocket transport close

leszekhanusz:
Should fix issue #130 

Also adding more debug logs


KingDarBoja:
Apologies for not being able to review this PR, been busy these weeks 😢

Issue #134, opened 2020-08-27
RuntimeError "There is no current event loop in thread" if not on main Thread


redreceipt:
I'm getting an error using the basic example syntax. You mention a warning below on the README and if this is it, I apologize, but I'm not familiar enough with AsyncIO to connect the dots. 

My code:
```py
class GraphQLClient:

    transport = AIOHTTPTransport(url="https://fake.com/graphql")
    client = Client(transport=transport, fetch_schema_from_transport=True)

    def request(self, body):
        return self.client.execute(body)

client = GraphQLClient()
r = client.request(
    gql(
    """
    query getUserID {
      login( email_or_phone_or_username: "{username}" password:"{pw}") {
        user_id
      }
    }
    """
    )
 )

```

The error:
```
  File "/Users/michael/Documents/Projects/kpffl/sleeper.py", line 16, in request
    return self.client.execute(body)
  File "/Users/michael/Documents/Projects/kpffl/venv/lib/python3.8/site-packages/gql/client.py", line 113, in execute
    loop = asyncio.get_event_loop()
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/events.py", line 639, in get_event_loop
    raise RuntimeError('There is no current event loop in thread %r.'
RuntimeError: There is no current event loop in thread 'Thread-8'.
```

leszekhanusz:
> You mention a warning below on the README and if this is it

No, it is something else. Your code is working for me.

Are you on a Mac ?
Are you using this code inside a new Thread ?

leszekhanusz:
Could you try to add before:

```python
import asyncio
asyncio.set_event_loop(asyncio.new_event_loop())
```

redreceipt:
That may have worked. Now I'm getting a different error:

`TypeError: Invalid or incomplete introspection result. Ensure that you are passing the 'data' attribute of an introspection response and no 'errors' were returned alongside: None.`

I'm running inside a Flask instance which I believe is multi-threaded, could that be why I needed to add the asyncio line?

leszekhanusz:
> I'm running inside a Flask instance which I believe is multi-threaded, could that be why I needed to add the asyncio line?

Yes, we should fix this.

For your next error, it means that the introspection query to fetch the schema failed. I will need more info.
Could you please try fetch_schema_from_transport=False ?

redreceipt:
I modified slightly with this code:

```py
[...]

PR #135, opened 2020-08-27
Fix running execute and subscribe of client in a new Thread

leszekhanusz:
In a new Thread, there is no default asyncio event loop like the main thread
And if we do asyncio.get_event_loop() OR if we create an asyncio.Event()
Then a RuntimeError will be generated

This commit fixes this problem.
This should fix the issue #134 and allow to use gql inside Flask which is multiThreaded


KingDarBoja:
That was a fast pull request, I will review this night :)

PR #139, opened 2020-09-22
Add option to send parameters in gql-cli

apple314159:
Example:
```console
echo 'query getContinent($code:ID!) { continent(code:$code) { name } }' | gql-cli https://countries.trevorblades.com --params code:AF
```


leszekhanusz:
Thanks for your PR.

I made some changes.

Note that for now it only works for **String** parameters I think. How could we make it work for **Int** ?

KingDarBoja (approved):
LGTM 🚀

KingDarBoja:
Why you closed this PR @apple314159 ?

PR #151, opened 2020-10-13
DSL: Allow to use GraphQL keys named type

### LINKED ISSUE #150: DSL not allowing to use a key named "type"
### Description

When trying to obtain a key named `type` from a specific GraphQL type, it raises an error.

The problem could be because the `type` attribute is already assigned to `DSLType` in https://github.com/graphql-python/gql/blob/fce2782e605bb2a4bfbbb9d6bf03ed4b459f1353/gql/dsl.py#L51-L53

### Steps to reproduce

- Have a `.graphql` file with the schema. For example:

```graphql
type Query {
    my_restaurant: Restaurant
}

enum RestaurantType {
    italian
    indian
    fast_food
}

type Restaurant {
    type: RestaurantType!
    address: String
    name: String
}
```

- Create a query to fetch `Restaurant.type`

```python
from schema import schema
from gql import Client
from gql.dsl import DSLSchema

client = Client(schema=schema)
ds = DSLSchema(client)
query = ds.Query.my_restaurant.select(
    ds.Restaurant.type
)
```

- It raises the following error:

```sh
Error
  File "/opt/project/src/apps/graphql/tests/test_graphql.py", line 80, in test_restaurant
    query = ds.Query.my_restaurant.select(
  File "/usr/local/lib/python3.8/site-packages/gql/dsl.py", line 91, in select
    selections=FrozenList(added_selections)
  File "/usr/local/lib/python3.8/site-packages/gql/dsl.py", line 73, in selections
    yield selection_field(_field).ast
  File "/usr/local/lib/python3.8/site-packages/gql/dsl.py", line 127, in selection_field
    raise TypeError(f'Received incompatible query field: "{field}".')
TypeError: Received incompatible query field: "Restaurant".
```

gdalmau:
Fixes #150

This is a PR trying to fix #150

### DIFF
--- a/gql/dsl.py
+++ b/gql/dsl.py
@@ -50,7 +50,7 @@ class DSLSchema(object):
 
 class DSLType(object):
     def __init__(self, type_):
-        self.type = type_
+        self._type = type_
 
     def __getattr__(self, name):
         formatted_name, field_def = self.get_field(name)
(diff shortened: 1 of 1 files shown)


leszekhanusz (approved):

gdalmau:
> Nice catch, in fact, graphql-core does use `type_` for internals ;)

@KingDarBoja should I rename it to `type_`?

leszekhanusz:
This is a variable only used in the class no ? So it should be fine as a private variable.
It could stay as `_type` for me.

KingDarBoja (approved):
No need to rename, just making a point of the usage of prefix and suffix for internals, everything looks great to me 🚀

gdalmau:
I tried the example in the issue linked #150 and it works with this version, so it's ready to merge 😄

PR #157, opened 2020-10-24
Feature aiohttp and requests transports logging

leszekhanusz:
Add logging messages for the aiohttp and requests transport.

Also add a logging page in the docs

### DIFF
--- a/docs/advanced/index.rst
+++ b/docs/advanced/index.rst
@@ -5,5 +5,6 @@ Advanced
    :maxdepth: 2
 
    async_advanced_usage
+   logging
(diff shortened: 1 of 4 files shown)


KingDarBoja (approved):
Everything looks great but should `RequestsHTTPTransport` have logging features as well? If that's the case, it can be provided on another PR. 🚀

leszekhanusz:
> Everything looks great but should `RequestsHTTPTransport` have logging features as well? If that's the case, it can be provided on another PR. rocket

I added it in this one.

Issue #147, opened 2020-10-10
Optional dependencies for transports


leszekhanusz:
Currently, the latest version installs all the dependencies necessary for all the underlying transports:

* requests
* websockets
* aiohttp

It should be possible to install gql without all the dependencies if we only use one transport!

My first idea was to try that by default, all the dependencies are added, but that it could be possible to remove dependencies if you only need one transport.

Apparently [this is not possible](https://github.com/pypa/setuptools/issues/1139)

Please comment if you have an idea.

nihaals:
Why not make it required to include an extra when installing the package? I don't think this is necessarily a bad thing and you would know you are missing a dependency when you attempted to import a transport (each transport should probably be separated into their own file to make sure the transport-only dependencies are only required for their class). In many cases users could even use transports like `requests` without including the extra as it's such a common dependency (although it should be documented that using the extra is still recommended in case another dependency is added for a specific transport)

leszekhanusz:
OK, that's what we are going to do. I made this issue to see if someone had a better idea for this.

leszekhanusz:
Implemented in release v3.0.0a4

PR #158, opened 2020-10-25
Optional dependencies for transports

leszekhanusz:
See issue #147

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,25 @@ jobs:
         env:
           TOXENV: ${{ matrix.toxenv }}
 
+  single_extra:
+    runs-on: ubuntu-latest
+    strategy:
+      dependency: ["aiohttp", "requests", "websockets"]
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install dependencies with only ${{ matrix.dependency }} extra dependency
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[${{ matrix.dependency }},test_no_transport]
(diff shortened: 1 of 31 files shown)


leszekhanusz:
Of course we'll need to document this before it is merged as it is a breaking change.

leszekhanusz:
I'm starting to think that my PR #112 was actually a bad idea. We could still revert it as we are still in alpha?

KingDarBoja:
Seems to be the same setup as I did at [graphql-server](https://github.com/graphql-python/graphql-server/blob/master/setup.py) but I am curious about the usage of [with supress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) statement in order to keep running the program if the user has not installed any "required" extras.

Regarding PR #112, well since you want to alert users to install required packages in order to use specific transports, the export should be keep on the same module scope, in other words, with this change it should avoid including all transports on global export (`__init__.py`).

leszekhanusz:
> Seems to be the same setup as I did at [graphql-server](https://github.com/graphql-python/graphql-server/blob/master/setup.py)

That's because I copied it from there :laughing:

KingDarBoja:
@leszekhanusz are you going to revert the #112 on this PR so it is two birds, one shot? 👓

leszekhanusz:
> @leszekhanusz are you going to revert the #112 on this PR so it is two birds, one shot?

Yes

leszekhanusz:
I modified the tests so that you can run tests with only a single transport dependency.
To do that:
* create a new python environment
* Install the dependencies for a single transport
* run the tests with the flag --aiohttp-only, --websockets-only or --requests-only depending on the selected dependency

to install test dependencies for only websockets for example:
```
pip install .[websockets,test_no_transport]
```

leszekhanusz:
Now I still need to add GitHub actions to test this automatically

PR #160, opened 2020-10-26
chore: stable Python 3.9 support

KingDarBoja:
GH Actions supports Python 3.9 as stable release. More info: https://github.com/actions/virtual-environments/issues/1740

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9-dev", "pypy3"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "pypy3"]
         os: [ubuntu-latest, windows-latest]
         exclude:
           - os: windows-latest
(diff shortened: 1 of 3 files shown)


leszekhanusz:
LGTM :rocket:

Actually the test with 3.9-dev never worked on my machine using tox (with conda) on my machine and this change does not solve this.

I could finally test it by adding:
```
conda_channels =                                                                                                        
    conda-forge
```
in the `[testenv]` section of tox.ini

Maybe we should add this to the CONTRIBUTING.md file

During testing I discovered that there are deprecation warnings under python3.9 and created issue #161

leszekhanusz (approved):

KingDarBoja:
@leszekhanusz looks like it is still not available on default, but it does on conda-forge or anaconda channels: https://anaconda.org/anaconda/python

> Maybe we should add this to the CONTRIBUTING.md file
Yeah, while the Python version is supported by default, I will add it to the `tox.ini` file in this PR.

> During testing I discovered that there are deprecation warnings under python3.9 and created issue
I ignored those while running py3.9-dev locally 😆 but glad you made a PR for that.

PR #156, opened 2020-10-24
Update aiohttp from 3.6.2 to 3.7.0

leszekhanusz:

PR #166, opened 2020-11-05
Update aiohttp from ==3.7.1 to >=3.7.1

bertonha:
yarl restrictions to version 1.6.0 was fixed on aiohttp 3.7.0


setup.py
```diff
@@ -4,7 +4,7 @@
 
 install_requires = [
     "graphql-core>=3.1,<3.2",
-    "yarl>=1.4,<1.6",
+    "yarl>=1.4,<2.0",
```
KingDarBoja, line 7: Actually it should be `yarl>=1.6,<2.0` based on https://github.com/aio-libs/aiohttp/pull/4973

setup.py
```diff
@@ -32,7 +32,7 @@
 ] + tests_requires
 
 install_aiohttp_requires = [
-    "aiohttp==3.7.1",
+    "aiohttp==3.7.2",
```
KingDarBoja, line 35: I think the change should be `aiohttp>=3.7.0` as you mentioned this changed was fixed at `3.7.0`.

KingDarBoja (changes requested):

setup.py
leszekhanusz: > I think the change should be `aiohttp>=3.7.0` as you mentioned this changed was fixed at `3.7.0`.

The 3.7.0 version is buggy (See PR #156) so this is probably a bad idea.
Should we request a specific version of aiohttp or should we use `aiohttp>=3.7.1` ???

KingDarBoja: Ah yes, there was a different bug regarding `typing-extensions` as well.

If that's the case, we must use a range for aiohttp like as `aiohttp>=3.7.1,<3.8.0` so any patch version can be used with this library and also avoid any breaking change from a minor release.

bertonha: @KingDarBoja I think all request changes was already applied, let me know if i miss something.

KingDarBoja (approved):
LGTM 🚀

PR #167, opened 2020-11-06
REFACTOR fetch schema at start of session

leszekhanusz:
Until now, the schema was fetched using an introspection query:
 - in the `__init__` of the client for sync transports
 - just before the first execute for the async transports

With this PR, both cases are modified so that the introspection query is executed just after the establishment of the first session.

This is a breaking changes for users using the RequestsHTTPTransport but only if they needed to use the schema before doing any query.


KingDarBoja:
Does this change improves overall performance? Seems like it does as now the schema will be fetched if needed and not everytime a request is made :thinking:

leszekhanusz:
It does improve performance (especially if we are using https) for the RequestsHTTPTransport as it will now use the same session for fetching the schema and for the first query.
For the async transports, it is basically the same.
But overall I think it is better to have the same logic for sync and async transports as it is more coherent.

Issue #138, opened 2020-09-08
How to reuse connection when using DSL?


Jeff17Robbins:
Each DSL operation calls `self.client.execute` which apparently creates a new connection each time.  Passing a `session` to `DSLSchema` doesn't work and causes an infinite recursion.

```
Traceback (most recent call last):
  File "gql_test2.py", line 45, in <module>
    query = ds.Query.listCasess().select(ds.ModelCasesConnection.items.select(ds.Cases.id))
  File "C:\Program Files\Python38\lib\site-packages\gql\dsl.py", line 38, in __getattr__
    type_def = self.schema.get_type(name)
  File "C:\Program Files\Python38\lib\site-packages\gql\dsl.py", line 38, in __getattr__
    type_def = self.schema.get_type(name)
  File "C:\Program Files\Python38\lib\site-packages\gql\dsl.py", line 38, in __getattr__
    type_def = self.schema.get_type(name)
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

Is there some known good way to reuse the connection when using DSL?  Is the `session` meant to work, and this is simply a bug?

leszekhanusz:
This is not possible to reuse the connection with the current DSL code.
The DSL code was made before all the 3.0 changes and is still quite experimental.
A refactor is probably needed so that it only returns a `DocumentNode` instead of needing to execute the query itself, this way we could execute the query how we want (in a session for example)

Jeff17Robbins:
That makes sense.  For now, I'm doing this:

Jeff17Robbins:
That makes sense.  For now, I'm doing this

```
    query = f'mutation {{ {str(_query)} }}' # where _query is built using the DSL
    res = session.execute(gql(query))
```

If I understood you properly, in the future, it might be possible to do this

```
    res = session.execute(_query)
```

and avoid the trip back to text and parsing?

leszekhanusz:
> If I understood you properly, in the future, it might be possible to do this
> 
> ```
>     res = session.execute(_query)
> ```
> 
> and avoid the trip back to text and parsing?

Yes, exactly, something like this.

leszekhanusz:
I refactored the DSL code and the proposed way is a little bit different.
Example:
```python
query = dsl_gql(ds.Query.hero.select(ds.Character.name))                                                            
result = session.execute(query)
```

leszekhanusz:
Fixed in v3.0.0a5

Issue #82, opened 2020-05-15
Document the "dsl" module/language


Cito:
The somewhat obscure "dsl" module/language included in gql should be documented. It is currently unclear to me where it is coming from, whether it mimics an existing thing or was an invention of @syrusakbary. Does anybody know?

Note: I'm not talking about the "schema definition language" of GraphQL here, but about the dsl module in gql. I think the name "dsl" is also a bit misleading because it usually means "domain specific language", but this is actually a general purpose language and not domain specific in the usual sense of "application domain specific". As far as I understand it is just an alternative way to express GraphQL documents in Python instead of JSON, Similar to the "SQL Expression language" in SQLAlchemy which allows writing SQL queries in Python.

Cito:
Btw, there is also an "improved-dsl" feature branch that has not yet been merged if anybody wants to look into this.

leszekhanusz:
FYI I added a very small example of the dsl module in the readme in PR #114

leszekhanusz:
Note: I have completely refactored the DSL code in PR #169 
It is a breaking change but:
- it should integrate correctly with the rest of the current code (with sessions)
- it supports subscriptions
- and it comes with full documentation :tada:

I've looked into the improved-dsl branch and it sounds quite obsolete except maybe for the
"Add alias to selection as keyword argument" commit which could be a good idea to simplify the aliases?

KingDarBoja:
> "Add alias to selection as keyword argument" commit which could be a good idea to simplify the aliases?

I am up for that change as well, maybe on same branch as refactored DSL 🤔

leszekhanusz:
Fixed in v3.0.0a5

PR #169, opened 2020-11-07
Refactor DSL code

leszekhanusz:
Refactoring of the DSL code.

- The DSLSchema only needs a schema, a Client is no more needed.
- new `dsl_gql` function to convert the DSL operations into a Document
- this allows DSL queries to be executed in a session instead of a client (Solves #138)
- Added typing annotations
- Added code documentation and sphinx docs (Solves #82)
- supports Subscriptions
- fix nested input arguments
- allow to set the alias directly in the select method
- allow multiple operations in a document
- allow to set the operation name

Simple example:
```
query = dsl_gql(DSLQuery(ds.Query.hero.select(ds.Character.name)))                                                            
result = client.execute(query) 
```


leszekhanusz:
I'll add documentation tomorrow before merging

gql/dsl.py
```diff
@@ -155,10 +155,10 @@ def __init__(self, graphql_type: GraphQLTypeWithFields):
             Don't instanciate this class yourself.
             Use attributes of the :class:`DSLSchema` instead.
 
-        :param graphql_type: a GraphQL type
+        :param graphql_type: ther GraphQL type definition from the schema
```
KingDarBoja, line 158: You made  typo here 👅

gql/dsl.py
```diff
@@ -1,179 +1,362 @@
...
+def dsl_gql(*fields: "DSLField") -> DocumentNode:
+    """Given arguments of type :class:`DSLField` containing GraphQL requests,
+    generate a Document which can be executed later in a
+    gql client or a gql session.
+
+    Similar to the :func:`gql.gql` function but instead of parsing a python
+    string to describe the request, we are using requests which have been generated
+    dynamically using instances of :class:`DSLField` which have been generated
+    by instances of :class:`DSLType` which themselves have been generated from
+    a :class:`DSLSchema` class.
```
KingDarBoja, line 52: I read too many `have been generated` phrases here, maybe something like:

> we are using requests which have been generated dynamically using instances of :class:`DSLField`, generated by instances of :class:`DSLType` which themselves originated from a :class:`DSLSchema` class.

gql/dsl.py
```diff
@@ -1,179 +1,362 @@
...
+    """
+
+    # Check that we receive only arguments of type DSLField
+    # And that they are a root type
+    for field in fields:
+        if not isinstance(field, DSLField):
+            raise TypeError(
+                f"fields must be instances of DSLField. Received type: {type(field)}"
+            )
+        assert field.type_name in ["Query", "Mutation", "Subscription"], (
```
KingDarBoja, line 78: Shouldn't this assertion match the enum types of the spec for [OperationType](https://spec.graphql.org/draft/#OperationType)?

Also graphql-core has [OperationType](https://github.com/graphql-python/graphql-core/blob/df8b5484a9e3ed69b1dcad66e923f9d1fea2cbdc/src/graphql/language/ast.py#L210) as an Enum.

KingDarBoja (commented):
Outstanding work @leszekhanusz 

Also it seems you also fixed #104, right?

gql/dsl.py
leszekhanusz: Well spotted!

gql/dsl.py
leszekhanusz: Good idea

leszekhanusz:
> Outstanding work @leszekhanusz

Thanks :smile: 

> Also it seems you also fixed #104, right?

No, this is not related, #104 is not using the DSL module at all.

gql/dsl.py
```diff
@@ -1,179 +1,364 @@
...
+                          for this field.
+        """
+
         added_args = []
         for name, value in kwargs.items():
             arg = self.field.args.get(name)
             if not arg:
                 raise KeyError(f"Argument {name} does not exist in {self.field}.")
-            arg_type_serializer = get_arg_serializer(arg.type, known_serializers=dict())
+            arg_type_serializer = self._get_arg_serializer(arg.type)
```
KingDarBoja, line 310: Unrelated but I always see this `_get_arg_serializer` requires `GraphQLInputType` but this type does not include `GraphQLInputField` type.

Seems like `graphql-core` always defines it as below:

```python
GraphQLInputType = Union[
    GraphQLScalarType, GraphQLEnumType, GraphQLInputObjectType, GraphQLWrappingType
]
```

@Cito this type definition seems to be based on https://spec.graphql.org/draft/#sec-Input-Objects.Type-Validation, right?

KingDarBoja (approved):
~Everything looks great, although I have made the alias change to work locally and tested it without any issue, should I push the commit to this branch so everything gets merged on one shot?~

Done 🍰

gql/dsl.py
Cito: GraphQLInputTypes are those that may be used as input types for arguments and directives. It's based on the same type in GraphQL.js which is based on the spec you linked.

gql/dsl.py
```diff
@@ -1,179 +1,378 @@
...
+            inner_serializer = self._get_arg_serializer(arg_type.of_type)
+            return lambda list_values: ListValueNode(
+                values=FrozenList(inner_serializer(v) for v in list_values)
+            )
+        elif isinstance(arg_type, GraphQLEnumType):
+            return lambda value: EnumValueNode(
+                value=cast(GraphQLEnumType, arg_type).serialize(value)
+            )
 
+        return lambda value: ast_from_value(
```
Cito, line 362: Maybe similar as above: `elif isinstance(arg_type, GraphQLScalarType): ... else: throw error`

gql/dsl.py
```diff
@@ -1,179 +1,378 @@
...
 from graphql.pyutils import FrozenList
 
 from .utils import to_camel_case
 
+log = logging.getLogger(__name__)
 
-class DSLSchema(object):
-    def __init__(self, client):
-        self.client = client
+GraphQLTypeWithFields = Union[GraphQLObjectType, GraphQLInterfaceType]
```
Cito, line 39: I would not create an alias for this, but use it inline. Also not that input types also have fields, so the name is misleading.

gql/dsl.py
```diff
@@ -1,179 +1,378 @@
...
-        return self.execute(query(*args, **kwargs))
 
-    def mutate(self, *args, **kwargs):
-        return self.query(*args, operation="mutation", **kwargs)
+class DSLType:
+    """The DSLType represents a GraphQL type for the DSL code.
+
+    It can be a root type (Query, Mutation or Subscription).
+    Or it can be an interface type (Character in the StarWars schema).
+    Or it can be an object type (Human in the StarWars schema).
```
Cito, line 143: The root types are also object types, so maybe rephrase like this:

It can be a root type (Query, Mutation or Subscription).
Or it can be any other object type (Human in the StarWars schema).
Or it can be an interface type (Character in the StarWars schema).

gql/dsl.py
leszekhanusz: The problem is that it is impossible in theory that the error happens if `_get_arg_serializer` is called by the `args` method.
So it means we have to add a meaningless test for this in order for the raise error line to get covered by the tests.
I propose instead to use `assert isinstance(arg_type, GraphQLScalarType)` here.

gql/dsl.py
leszekhanusz: Indeed, it is better. Done

gql/dsl.py
leszekhanusz: Done

gql/dsl.py
leszekhanusz: You are right @KingDarBoja , there was a problem here because the `_get_arg_serializer` method was called with a _GraphQLField_ argument instead of a type if the parent input type was a _GraphQLInputObjectType_.
I modified the code in my last commit to get the type directly instead of using the field so that now the typing info should be correct I think.

gql/dsl.py
Cito: Right, an assert may be appropriate in such a situation where you have a condition that should theoretically *never* fail. Note that it will be stripped away with "-O" which can be seen as an advantage (better performance in production) or disadvantage ("impossible" but still happening errors in production not detected). If you use asserts, then you should do it consistently for such cases in the whole codebase, (see also [here](https://github.com/graphql-python/graphql-core-legacy/issues/244)).

One of the reasons to add an "if isinstance" or an "assert" is that mypy should (in theory) understand that so that you don't need the "cast" (not sure if that works in the latest version of mypy). It is also a hint to the reader of the code. Btw, GraphQL-Core also has an `assert_scalar_type` function that you can use here. It raises a TypeError though instead of an AssertionError. Its return value is GraphQLScalarType so that should also help mypy.

I also add a "# pragma: no cover" to not decrease coverage for such code that should in theory never execute.

leszekhanusz:
I can't believe it :man_facepalming: , the `_get_arg_serializer` function was not needed after all as all it did was to rewrite the `ast_from_value` method. It is now removed.

gql/dsl.py
leszekhanusz: See my last commit, this method was removed so this problem as well :laughing:

gql/dsl.py
leszekhanusz: all this is irrelevant now that the _get_arg_serializer method was removed :rofl:

KingDarBoja (approved):
Perfect, I loved the latest changes and the operation name feature :D

Issue #178, opened 2020-12-05
Logging is too verbose


tylermenezes:
Right now all websocket messages are being sent at `INFO` level, and there's no way to change it for graphql-python (only globally). Useful information is often emitted at `INFO` level, but it becomes hard to read with all the websocket traffic. Especially when it fetches the initial schema (which in my case is large enough that it sometimes crashes my IDE).

It would be helpful if either:

1. Websocket traffic is emitted at `DEBUG` (which seems reasonable to me, as they're only needed when debugging).
2. Logging levels could be overriden just for this package.

I'm happy to make the pull request if you're open to one or the other option.

leszekhanusz:
What I wanted was to have a single level used only for the messages exchanged between the frontend and the backend, and a lower level for more even lower debug messages.

Of course this could be changed if necessary.

Please note that with the python logging module it should be possible to configure different log levels for different loggers and also for different handlers so that you could ignore the websockets logs ([See logging control flow](https://docs.python.org/3/howto/logging.html#logging-flow))

alexisrolland:
I also find the logging quite verbose and a bit difficult to read. When the query sent to the API is logged it contains a lot of new lines and spaces which damage the readibility. It is not the case in the response which I find more readable. Example:

```
2020-12-12 09:02:07,757 - gql.transport.requests - INFO - >>> {"query": "mutation updateJobById($jobId: Int!, $jobPatch: JobPatch!) {\n  updateJobById(input: {jobPatch: $jobPatch, id: $jobId}) {\n    job {\n      id\n      title\n      status\n      progress\n      step\n      spaceId\n      configurationId\n    }\n  }\n}\n", "variables": {"jobId": 1, "jobPatch": {"status": "running"}}}
2020-12-12 09:02:07,796 - gql.transport.requests - INFO - <<< {"data":{"updateJobById":{"job":{"id":1,"title":"test","status":"running","progress":"1.00","step":1,"spaceId":1,"configurationId":1}}}}
```

leszekhanusz:
Here is a small example on how to disable the logs (for the requests transport but it should work the same way for other transports)

It works by getting the logger from the transport and changing the level for this logger only (so that all the other logs in your code are not affected).

I'll make a PR to improve the documentation about this.

```python
import logging                                                                                                          
                                                                                                                        
logging.basicConfig(level=logging.INFO)                                                                                 
                                                                                                                        
from gql import Client, gql                                                                                             
[...]

PR #183, opened 2020-12-12
DOC add documentation on how to disable the transport logs

leszekhanusz:
Should help with issue #178

### DIFF
--- a/docs/advanced/logging.rst
+++ b/docs/advanced/logging.rst
@@ -18,4 +18,31 @@ For even more logs, you can set the loglevel at **DEBUG**:
     import logging
     logging.basicConfig(level=logging.DEBUG)
 
+Disabling logs
+--------------
+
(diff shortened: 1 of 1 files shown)


KingDarBoja (approved):
Excellent. Regarding the following statement:

> When the query sent to the API is logged it contains a lot of new lines and spaces which damage the readibility. 

At the linked issue, would be good to point out that this is on purpose as GraphQL schemas and queries have this formatting in order to be passed into the Parser/Lexer.

Maybe they expect a pretty printing of those JSON responses, which can be achieved very easily by using something like [graphql-server](https://github.com/graphql-python/graphql-server/blob/f5e8302d1320b013b441844df059e90ae83d04a0/graphql_server/__init__.py#L130) does:

```python
import json

json.dumps(data, indent=2, separators=(",", ": "))
```

Issue #188, opened 2021-01-23
Customizable ExecutionResult, or how to get other API response parts


extrin:
https://github.com/graphql-python/gql/blob/459d5ebacce95c817effcd18be152d6cf360cf2b/gql/transport/requests.py#L174
### Case:
The GraphQL API that I need to make requests to, responds with some other useful information, except of `data` block:
```json
{
    "data": {...},
    "extensions": {
        "pageInfo": {
            "limitPage": 200,
            "totalCount": 23306516,
            "hasNextPage": true,
            "lastId": 41922710
        }
    }
}
```
But, as **gql** transport composes `ExecutionResult` from `data` and `errors` blocks only, I'm not able to get and use the `extensions` part in my code. 
I was hoping to use `hasNextPage` and `lastId` values to iterate through pages of data, to load all the data to database. 
### Question:
Is there any way to get `extensions` part of response, along with `data` and `errors`, using **gql**?

leszekhanusz:
Currently this is not possible, the `execute` and `subscribe` methods of the session only return the data or throw an error and as you noticed, the ExecutionResult instance is not created with the possible extensions field in the transports.

If you want direct access to the ExecutionResult instance, you could use the `_execute` and `_subscribe` methods of the session.

A simple PR could be made to add the extensions field of the ExecutionResult object for all the transports.

KingDarBoja:
AFAIK the execution result only returns the `data` and `errors` fields according to the official spec and js implementation: https://graphql.org/graphql-js/execution/#execute

So no clue how that graphql-API is returning such extensions.

extrin:
Thanks for your answers!
I see that the extensions field [was added](https://github.com/graphql-python/graphql-core/commit/3359fe60d9940b174270426438a89d629b6cc6e2) to ExecutionResult in graphql-core. 
It seems that this change is not in release version yet. I just tried to fork repo and add extensions field for transports, but the ExecutionResult imported from graphql doesn't have this field either.

dkbarn:
> A simple PR could be made to add the extensions field of the ExecutionResult object for all the transports.

Even if "extensions" was added to the ExecutionResult object for all transports, it would not be retrievable using Client.execute, because that method returns a Dict object containing only the contents of the "data" key: https://github.com/graphql-python/gql/blob/master/gql/client.py#L420

KingDarBoja:
This should be fixed by #190

leszekhanusz:
[...]

PR #190, opened 2021-02-09
Add extensions field to ExecutionResult (#188)

extrin:
Add the extensions field of the ExecutionResult object for all transports.

### DIFF
--- a/gql/transport/aiohttp.py
+++ b/gql/transport/aiohttp.py
@@ -228,11 +228,7 @@ class AIOHTTPTransport(AsyncTransport):
                     f"Server did not return a GraphQL result: {result_text}"
                 )
 
-            if (
-                "errors" not in result
-                and "data" not in result
-                and "extensions" not in result
-            ):
+            if "errors" not in result and "data" not in result:
                 result_text = await resp.text()
                 raise TransportProtocolError(
                     "Server did not return a GraphQL result: "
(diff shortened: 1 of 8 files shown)


gql/transport/aiohttp.py
```diff
@@ -228,15 +228,23 @@ async def execute(
                     f"Server did not return a GraphQL result: {result_text}"
                 )
 
-            if "errors" not in result and "data" not in result:
+            if (
```
leszekhanusz, line 231: I am not sure we need to modify this condition.

[https://spec.graphql.org/draft/#sec-Response-Format](https://spec.graphql.org/draft/#sec-Response-Format)

> 
> If the operation encountered any errors, the response map must contain an entry with key errors. The value of this entry is described in the “Errors” section. If the operation completed without encountering any errors, this entry must not be present.
> 
> If the operation included execution, the response map must contain an entry with key data. The value of this entry is described in the “Data” section. If the operation failed before execution, due to a syntax error, missing information, or validation error, this entry must not be present.
> 
> The response map may also contain an entry with key extensions. This entry, if set, must have a map as its value. This entry is reserved for implementors to extend the protocol however they see fit, and hence there are no additional restrictions on its contents.

It is not clear for me if it is possible to receive a response with only the key "extensions"

leszekhanusz:
Sorry for the delay.
Could you please add some tests for this functionality?
Thanks

leszekhanusz (approved):

Issue #200, opened 2021-04-06
Any plan for handling "ka" (keep-alive) and ping-pong checks?


sneko:
Hi @leszekhanusz ,

I'm wondering if there is any plan to support the keep-alive message? It comes from the server to make sure the connection is still alive, and if the client receives no KA within a specific timeout defined, it behaves as the connection was lost.

When seeing:
https://github.com/graphql-python/gql/blob/d7d871f89436a5cda05fda89319476117dbb73ca/gql/transport/websockets.py#L186
and:
https://github.com/graphql-python/gql/blob/d7d871f89436a5cda05fda89319476117dbb73ca/gql/transport/websockets.py#L316

I understand that's not implemented, but just wondering if it's in the pipe.

Also, in the same kind of standards, Keep-Alive makes sure the client will react in case of lost connection, but on it's end the server won't know if the connection was lost (router shut down for example, or loss of connectivity, that are not proper terminations). This is usually handled for websockets by an end-to-end "ping-pong" protocol between the client and the server:
* The server sends a "ping" every X interval
* The client has to reply "pong" within a timeout defined otherwise the server considers the connection as lost (which help freeing memory and doing specific stuff like updating client status in the database...)

Is it also something on your plan? It would definitely help.

Thank you,

sneko:
@leszekhanusz I'm almost done with implementing the keep-alive part. I will do a PR tomorrow.

I hope to finish receiving a ping from the server and sending back the pong.

Also, I'm not that familiar with Python tests and I'm not sure how to proceed. For now I'm testing with real software... After PRs I would be glad if you could take a look to see how new tests could fit the commits.

leszekhanusz:
Hi @sneko

You are right, it could be a good idea to implement closing the connection if we don't receive a keep-alive message.
We could add a `keep_alive_timeout` argument to the `__init__` method of `WebsocketsTransport`which is None by default (Keep Alives are just ignored as it is now) and if we don't receive a keep alive during this time raise a TimeoutError.

On the other hand, the client should never send something in response because the `WebsocketsTransport` transport should follow the [specification](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md)

Any deviation from this protocol should be done in a custom class created by the user.

sneko:
> On the other hand, the client should never send something in response because the WebsocketsTransport transport should follow the specification

@leszekhanusz I guess for the second paragraph you meant about the `ping-pong` pattern.

[...]

PR #201, opened 2021-04-07
Handle keep-alive behavior to close the connection

sneko:
Hi @leszekhanusz ,

Here the PR for the keep-alive behavior https://github.com/graphql-python/gql/issues/200, tested with real servers sending `ka` messages.

I read the CONTRIBUTING.md but struggled with the `make check`, I installed them but it changes all my files (including yours) like if parameters were not taken in account. Moreover `isort` tells me some lines are >88chars but **no** other tools automatically adjust them (I thought it would?). I guess I missed something... I did the venv creation + `make dev-setup`.

_Tip for others: in my app code I forgot to `await ws_transport.wait_close()` making the reconnection after a keep-alive failure failing because some stuff were not clean in the meantime (`asyncio` giving the priority to other tasks including mines trying a reconnect)._

Tell me if I need to adjust something :)

### DIFF
--- a/gql/transport/websockets.py
+++ b/gql/transport/websockets.py
@@ -1,4 +1,5 @@
 import asyncio
+from contextlib import suppress
 import json
 import logging
 from ssl import SSLContext
@@ -94,6 +95,7 @@ class WebsocketsTransport(AsyncTransport):
         connect_timeout: int = 10,
         close_timeout: int = 10,
         ack_timeout: int = 10,
+        keep_alive_timeout: int = 0,
         connect_args: Dict[str, Any] = {},
     ) -> None:
         """Initialize the transport with the given parameters.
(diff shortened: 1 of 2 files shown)


sneko:
@leszekhanusz I fixed the virtualenv so it's not longer modifying lot of things.

Note: from https://github.com/graphql-python/gql/blame/master/CONTRIBUTING.md#L34
it should be instead writing as inside your `Makefile` with quotes: `python pip install -e ".[dev]"`
_by the way, in the `.md` it's about `dev` and in the `Makefile` about `[test]`, is it intended?_

Now I'm still getting the line length errors... I'm wondering which tool should be managing this.

leszekhanusz:
@sneko Could you please check if the refactor is working correctly for you ?

PR #210, opened 2021-05-21
Add support for variable definitions in dsl

wallee94:
Hi all,

I was working in a small project and I needed a way to compose queries dynamically with variable definitions, so I can create the queries once depending on initial settings and just change the variables when sending them.
I created a `DSLVariableDefinitions` and `DSLVariable` class in the dsl module to add the option. I tried to make it similar to the current API :slightly_smiling_face: 

This is an example of how it works:

```python
var = DSLVariableDefinitions()
op = DSLMutation(
    ds.Mutation.createReview.args(review=var.review, episode=var.episode).select(
        ds.Review.stars, ds.Review.commentary
    )
)
op.variable_definitions = var
query = dsl_gql(op)
```

And it generates:

```graphql
mutation ($review: ReviewInput, $episode: Episode) {
  createReview(review: $review, episode: $episode) {
    stars
    commentary
  }
}
```

I have no experience with sphinx, please let me know if I need to adjust something.


leszekhanusz:
Thanks! I'll look into it.

leszekhanusz:
Your solution does not work with variables in input objects like this:

```
createReview(                                                                                                         
    review: {stars: $stars, commentary: $commentary}                                                                    
    episode: $episode                                                                                                   
  )
```

To solve this, the only way is to modify the recursive `ast_from_value` function from graphql-core.
As this function cannot be modified directly in graphql-core because it is a one-to-one copy of `astFromValue` from graphql-js
I made a copy paste inside gql. I agree that this is quite hackish.

graphql-js itself is currently in the process of changing quite a few things (See [graphql-js PR #3049](https://github.com/graphql/graphql-js/pull/3049)) and changing the `astFromValue` function to `ValueToLiteral`

I'll make an issue on graphql-js side to do this a little bit cleaner.

wallee94:
Oh, you're right! I missed that option.

I agree that the only workaround is going to be copying the `ast_from_value` function. It is hackish but at the same time I think it makes sense to have a custom version of `ast_from_value` in the dsl module, because returning `VariableNodes` sounds very specific to composing a query.
It looks like `valueToLiteral` will add support for a `valueToLiteral` method in `GraphQLInputType`, but that wouldn't help because `DSLVariables` should work with any type

leszekhanusz:
Looks good!

I added some docs.

Now for the hard part, could you please add tests so that the code coverage stays at 100% ?

wallee94:
Great! Thank you @leszekhanusz.
Sure, I'll work on the tests

wallee94:
I added the tests and I think coverage is back to 100% again 🙂

gql/dsl.py
```diff
@@ -194,6 +277,69 @@ class DSLSubscription(DSLOperation):
     operation_type = OperationType.SUBSCRIPTION
 
 
+class DSLVariable:
```
leszekhanusz, line 280: Please add typing hints for all methods

leszekhanusz (changes requested):
Great!

leszekhanusz (approved):

leszekhanusz:
LGTM

leszekhanusz:
Thanks again for your PR!

PR #214, opened 2021-05-24
Bump websockets to 9

leszekhanusz:


tests/conftest.py
```diff
@@ -125,7 +125,7 @@ def __init__(self, with_ssl: bool = False):
 
     async def start(self, handler):
 
-        import websockets
+        from websockets.legacy import server
```
aaugustin, line 128: You were doing `import websockets` then `websockets.server.serve`.

Generally speaking, importing module `a` doesn't make `a.b` automatically available in the namespace.

I'd recommend to change `import websockets` to `import websockets.server`, or to change `websockets.server.serve` to `websockets.serve`.

aaugustin: (I'm not a big fan of seeing the `legacy` private module leaking in third-party code.)

leszekhanusz: Thanks for your input. I made another PR to fix this: #217

PR #232, opened 2021-08-22
feat(transport/requests): RequestsHTTPTransport execute func add extra_args param

ma-pony:
RequestsHTTPTransport execute function support input headers.
like this
```
result = client.execute(
    query, extra_args={"headers": {"authorization": "xxx-123"}}
)
```

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -103,6 +103,7 @@ class RequestsHTTPTransport(Transport):
         variable_values: Optional[Dict[str, Any]] = None,
         operation_name: Optional[str] = None,
         timeout: Optional[int] = None,
+        extra_args: Dict[str, Any] = None,
     ) -> ExecutionResult:
         """Execute GraphQL query.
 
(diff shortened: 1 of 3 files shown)


leszekhanusz:
Are you aware that you can already set headers using `RequestHTTPTransport` using the `headers` argument of the transport ?

gql/transport/requests.py
```diff
@@ -145,7 +147,8 @@ def execute(  # type: ignore
 
         # Pass kwargs to requests post method
         post_args.update(self.kwargs)
-
+        if extra_args:
+            post_args.update(extra_args)
```
leszekhanusz, line 151: Please add two newlines, one before the if block and one after

leszekhanusz (changes requested):
LGTM except some minor cosmetic requests (newlines)

ma-pony:
> Are you aware that you can already set headers with `RequestHTTPTransport` using the `headers` argument of the transport ?

yes, I know it.
but my case like 
```
transport = RequestTransport(url=url)
client = Client(transport=transport)

user_a_token = client.execute(user_a_login_gql)
user_a_book = client.execute(user_a_book_gql, extra_args={"headers": {"authorization": "user_a_token"}})

user_b_token = client.execute(user_b_login_gql)
user_b_book = client.execute(user_b_book_gql, extra_args={"headers": {"authorization": "user_b_token"}})

...

```

gql/transport/requests.py
ma-pony: > Please add two newlines, one before the if block and one after
fixed

leszekhanusz (approved):

PR #234, opened 2021-08-25
fix(gql.client): update type to permit passing floats

ossareh:
The type hint for execute_timeout are incorrect; the call is ultimately passed to `asyncio.wait_for` which supports passing:

 * `int`; interpreted as seconds
 * `float`; interpreted  as `<seconds>.<milliseconds>`.
 * `None`; which results in blocking forever

This commit fixes the type hint and updates the docs to inform the reader what results from passing in `None`.

My main concern about this change is not being clear why it was chosen to specify seconds as though they're ints; it may be that this change now forever subscribes this API to being compatible with `asyncio.wait_for` - and it's not clear to me if that is your overall intent.


leszekhanusz:
Well spotted, we could indeed accept floats.
For the linting issues, you should install the correct dev dependencies with `pip install .[dev]`, then run `make check`

You may want to also modify the typing hints in websockets.py:
```python
        connect_timeout: int = 10,                                                                                      
        close_timeout: int = 10,                                                                                        
        ack_timeout: int = 10,                                                                                          
        keep_alive_timeout: Optional[int] = None
```

Issue #198, opened 2021-03-19
DSLQuery with fragments


Samolevar:
Hi. Is it possible to create query with dsl if I need to pass a fragment or inline fragment within query?
No examples for that, and search didn't help either.
For example I want to create something like this:
 query getNode {
          node(id: "1234") {
            id
            ... on Model {
              id
              name        
            }
          }
        }

leszekhanusz:
Sorry, this is not implemented for dsl.
I agree that it is really needed.

Issue #125, opened 2020-07-31
AWS Appsync support?


kingtut:
AWS Appsync seems to require an `extensions` field on the payload. Does the WebsocketsTransport support this?

https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html

```{
  "id": "eEXAMPLE-cf23-1234-5678-152EXAMPLE69",
  "payload": {
    "data": "{\"query\":\"subscription onCreateMessage {\\n onCreateMessage {\\n __typename\\n message\\n }\\n }\",\"variables\":{}}",
    "extensions": {
      "authorization": {
        "accept": "application/json, text/javascript",
        "content-type": "application/json; charset=UTF-8",
        "X-Amz-Security-Token": "XXX",
        "Authorization": "AWS4-HMAC-SHA256 Credential=XXXXXXXXXXXXXXXXXXXX/20200401/us-east-1/appsync/aws4_request, SignedHeaders=accept;content-encoding;content-type;host;x-amz-date;x-amz-security-token, Signature=b90131a61a7xxx",
        "content-encoding": "amz-1.0",
        "host": "example1234567890000.appsync-api.us-east-1.amazonaws.com",
        "x-amz-date": "20200401T001010Z"
      }
    }
  },
  "type": "start"
}
```

leszekhanusz:
It is not currently supported. Adding the `extensions` field is not enough, the protocol is a little bit different and we should accept the `start_ack` message which is not part of the current protocol for example. Do you know if a public server exist somewhere so that we could test a new implementation ?

KingDarBoja:
There seems to be a tutorial for testing this implementation using a boilerplate React App: https://aws.amazon.com/blogs/mobile/appsync-realtime/ and https://docs.amplify.aws/lib/graphqlapi/getting-started/q/platform/js#mocking-and-local-testing

But I couldn't find the protocol on their repository: https://github.com/aws/aws-appsync-community

Only this blog post kinda explain it besides the one shared by the issue author: https://aws.amazon.com/blogs/mobile/appsync-realtime/

joseph-wortmann:
I have this implemented in very alpha form. I have tested it with OIDC credentials and it works. I need to test it with APK Keys and IAM (sigv4) credentials and then I'll submit a pull request.

[...]

PR #235, opened 2021-08-26
DSL implementation of fragments

Cito:
This is my suggestion for #198

### DIFF
--- a/docs/advanced/dsl_module.rst
+++ b/docs/advanced/dsl_module.rst
@@ -276,7 +276,7 @@ Once your fragment is defined, to use it you should:
 
     query_with_fragment = DSLQuery(ds.Query.hero.select(name_and_appearances))
 
-* add it as an attribute of :func:`dsl_gql <gql.dsl.dsl_gql>` with your query::
+* add it as an argument of :func:`dsl_gql <gql.dsl.dsl_gql>` with your query::
 
     query = dsl_gql(name_and_appearances, query_with_fragment)
 
@@ -311,7 +311,7 @@ To define an `Inline Fragment`_, you have to:
 
     human_fragment.select(ds.Human.homePlanet)
 
-Once your fragment is defined, to use it you should:
+Once your inline fragment is defined, to use it you should:
 
 * select it as a field somewhere in your query::
 
--- a/docs/advanced/dsl_module.rst
+++ b/docs/advanced/dsl_module.rst
@@ -252,6 +252,93 @@ It is possible to create an Document with multiple operations::
         operation_name_3=DSLMutation( ... ),
     )
 
+Fragments
+^^^^^^^^^
+
+To define a `Fragment`_, you have to:
+
+* Instanciate a :class:`DSLFragment <gql.dsl.DSLFragment>` with a name::
+
+    name_and_appearances = DSLFragment("NameAndAppearances")
+
+* Provide the GraphQL type of the fragment with the
+  :meth:`on <gql.dsl.DSLFragment.on>` method::
+
+    name_and_appearances.on(ds.Character)
+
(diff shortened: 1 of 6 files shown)


leszekhanusz:
> Sorry, I must leave it to you or others to work out the details due to lack of time. I just wanted to provide a rough idea how this can be implemented.

Ok, no worries, I'll propose something.

chadfurman:
I also would be interested in helping if the work isn't all done at this point

chadfurman:
Checking in to avoid duplicate efforts.  I'm setting up my local env right now and going to read over the code in https://github.com/graphql-python/gql/commit/d505962abdb4795e76f994ef93a80b7e3fae8df4 plus try and get a basic inline fragment to run + run the tests + evaluate the above comment against the commit.

leszekhanusz:
@chadfurman 
Now that I've started working on this, I would like to finish it. I hope you don't mind.

I am currently looking at how we could also support normal fragments into this PR.

That's when I realized that the `DSLSelection` class was actually used for two different purposes:
 - for elements which can be arguments of the select method (selectable elements)
 - for elements on which we can use the select method (containers)

So I decided to split it into DSLSelectable and DSLSelector.

Next I am going to introduce the `FragmentSpreadNode` and `FragmentDefinitionNode` from graphql-core AST.

chadfurman:
Not a problem!  Ping me if you want help on anything -- this or another ticket   I'll switch tasks 

Thank you for your effort and time :)

leszekhanusz:
> Not a problem! Ping me if you want help on anything -- this or another ticket. For now, I'll switch tasks
> 
> Thank you for your effort and time :)
> 
> fwiw: I have my local env setup and all dev deps on the right version now w/ tests and `make check` all fine sweat_smile
> 
> You have a very solid project here. Great work.

Thanks! If you want something to do, you could maybe check the issue #125, diving into the Appsync protocol and creating another transport AWSAppsyncTransport which inherits WebsocketsTransport.

leszekhanusz:
>     * Not sure if the DSL should really expose different objects for inline and named fragments. The distinction could be actually left transparent to the user, e.g. the dsl_gql function could optimize automatically by using named fragments instead of inline fragments. For the users of the DSL it makes no difference, since they can create fragment objects once and reuse them anyway, without needing to explicitly create named fragments.

@Cito I tried to combine normal and inline fragments in a single class but they are really different beasts, with normal fragments which need a name and can have variables.
With my current code, if you use a `DSLFragment` in dsl_gql, you'll get a `FragmentDefinitionNode`, but if you use it as an argument of a select, then you get a `FragmentSpreadNode`

So now we need to know when we use a fragment as an argument in a `select` whether this is a normal fragment, and in this case we return a `FragmentSpreadNode` or whether it is an inline fragment and we return an `InlineFragmentNode`. If we used the same class, we could detect it somehow by checking if the fragment has a name, but it might get confusing and I think it is better that we stay explicit here.

Cito:
> with normal fragments which need a name and can have variables

But as far as I see, fragments with vars were experimental only (not part of the spec) and have now been [deprecated](https://github.com/graphql/graphql-js/pull/2920) in Graphql.js (and hence in GraphQL-core as well).

leszekhanusz:
Interesting... But the main problem remains and it feels cleaner to me to keep the classes separate.
What do you think of the current state of this PR? Is this something we could agree on?

Cito:
LGTM, it's definitely cleaner and allows more control over the generated query.

leszekhanusz:
Good! I'll add some documentation and merge this PR next weekend.

vedantpuri:
@leszekhanusz  Any updates on this ?

leszekhanusz (approved):

PR #239, opened 2021-09-13
Aws appsync websocket transport

chadfurman:
AWS AppSyncWebsocketTransport inspired by (and largely copied from) the work of @joseph-wortmann 

For original discussion thread, go here: https://github.com/graphql-python/gql/issues/125

At the time of creating this draft PR, these are the next steps:

1. Revisit the _send_query overload in AppSyncWebsocketTransport -- note that I'm not sure about "data" being passed to "get_headers" as I need to revisit what, exactly, is supposed to be getting signed and how, exactly, the non-IAM authorizations are including the request data.
2. ensure nothing critical is missing from the _send_query overload. It's substantially smaller than the method in WebsocketTransport so I'm 99.99999% certain I'm missing a substantial amount of logic there. I simply ran out of time to verify
3. Try to instantiate a connection. This may require revisiting the AWSRequest object. Specifically, there's a "AWSRequestPreparer" but honestly "auth.add_auth()" is likely sufficient I just need to make sure that I build the URL correctly, I think?
4. figure out what was lost by stripping out the new connect() and subscribe() methods and how to replace that. Do we need to add the hook in WebsocketTransport, as was suggested? Or can we take the alternative path and build the URL up-front and let everything else automatically happen?
5. Definitely need some code cleanup. I started with some tests, but the coverage is abysmal and I'm not even trying to pass black/flake8 etc.

Follow the discussion here below for further updates.


chadfurman:
picking this back up today.  Going to start with:
1. Read through the WebsocketTransport base class and list out key elements we need to replicate in _send_query.  Brainstorm if it's possible to re-use the base-class's _send_query somehow
2. Interface with / overload `connect()` and `subscribe()` as necessary.  See https://github.com/graphql-python/gql/issues/125#issuecomment-913187491
3. Start code cleanup.  Work on black and flake8 and, if time allows, increase test coverage.

chadfurman:
A couple questions, as I read through this:

1. Why does our "_send_query" have to wait for "ack" but the original websocket implementation does not?
2. Why does our "_send_query" need to have its own receive_data_loop()?  Wouldn't the one from the base class's `connect()` be sufficient?

Note w.r.t. the above two questions, I've removed the wait_for_ack and the new receive_data_loop from the overloaded _send_query().  I can easily add them back, but I wanted to be as compatible with the base _send_query as possible and did not want to leak extra event loops nor duplicate wait events.

Note also that, because we've overloaded the _send_query() method as suggested and because we're manipulating the URL once during init, we've left both the `subscribe()` and `connect()` methods out of this implementation.

Next steps: 
1. expand test coverage to the auth types as well as the _send_query method
2. clean up flake8 and black errors
3. add better docblocks
4. add instructions on expected usage
5. test connecting to AWS and subscribing to an appsync endpoint.  Trigger a mutation and ensure that we receive the modified data.  Follow the instructions for connecting and, if any errors occur, update the documentation and code as necessary.
6 explore non-realtime queries and mutations to appsync -- these likely do not go over `wss` and so we may need a controller class which uses AIOHttpRequest under the hood with an option to overload this.

Will be pushing code shortly.

leszekhanusz:
>     1. Why does our "_send_query" have to wait for "ack" but the original websocket implementation does not?

Because with the aws protocol, an ack is sent by the backend but it is not the case with the apollo protocol

>     2. Why does our "_send_query" need to have its own receive_data_loop()?  Wouldn't the one from the base class's `connect()` be sufficient?

Indeed, only the one in the connect is necessary

chadfurman:
Very little time this weekend.  Going to start on expanding the test coverage and trying to clean up the code checks.

chadfurman:
Started working on getting existing appsync tests to pass.  Found out that the `add_auth` call is actually failing due to missing credentials -- I expected the Signer would find them itself.  I'll have to grab the credentials somehow and bind them.  

Gotta stop here for the night, unfortunately.  Didn't get very far.  Will keep coming back to thins until I get it, though.

chadfurman:
Looking into how to get AWS credentials object into `add_auth`

chadfurman:
Figured out it was because my profile was wrong.  Need to add a test for NoCredentialsError() where I inform the user that they need to specify a profile or set their default profile in their ~/.aws/credentials file (not sure how this works in windows, but I assume it's similar)

chadfurman:
New error is about the fake request:

```

self = <botocore.auth.SigV4Auth object at 0x7fa69bba8220>
request = <botocore.awsrequest.AWSRequest object at 0x7fa69ab03b80>

    def credential_scope(self, request):
        scope = []
        scope.append(request.context['timestamp'][0:8])
        scope.append(self._region_name)
        scope.append(self._service_name)
        scope.append('aws4_request')
>       return '/'.join(scope)
E       TypeError: sequence item 1: expected str instance, NoneType found
```

I'm guessing I need to set a timestamp on the request or find a better way of creating it that will set my timestamp for me.  I had really hoped to avoid having to muck with dates.

chadfurman:
Adding the exception handling for the missing credentials and working on getting the last test to pass before adding more.

chadfurman:
Added the exception handling for the missing credentials.  Also added exception handling for missing region even if credentials are provided.  Got all tests to pass as-is.  Next will be adding the e2e tests that talk to AWS using my actual credentials, and then also tying it into the existing websocket server tests as well.  Then I'll do a smoke test and clean up the docblocks and add documentation.

leszekhanusz:
For the linting issues, you can run `make check` to fix most problems (See Contributing.md file)

For the failing tests, you should add an aws element in the `all_transport_dependencies` in conftest.py and mark your tests with it.

chadfurman:
Starting.

Goal:
* Add "aws element" as suggested for failing tests
* Run "make check" for linting issues
* review existing websocket tests
* devise how I may tie into the existing websocket tests
* review existing e2e tests
* devise how the aws transport e2e test should be structured
* Commit the beginning of my e2e test

chadfurman:
I appear to have had some sort of system failure.

chadfurman:
Reformatted and installed latest fedora, back up and running.  Not sure I'll be able to get much done tonight, unfortunately, but I should be able to add the "aws" line and fix the linting once I get pycharm and git set back up.

chadfurman:
Going to have to stop here.  Let's see what happens with CI.  I'll pick away a bit more throughout the week now that I'm back up and running.  No idea what happened, but my filesystem randomly became "readonly" in the middle of playing with PyCharm and tweaking some battery-related stuff.

chadfurman:
Not sure why the tests still failed.  Will look into that later.  Looking into the existing websocket test and going to try to get it to validate my aws extension as well.  Not sure if I'll need to copy/paste lots of it (I'd prefer not to).  Need to better understand that websocket server thing you got in conftest.py

chadfurman:
**A few learnings from test_websocket_exceptions.py**
* Cassette / VCR seems to be a way to manage fixtures
* Many tests are async
* WebsocketServerHelper can send acks
* these tests rely on passing in a server and in some cases a client_and_server
* I'll have to refactor this test suite to allow me to inject a transport, and then create a test runner that wraps this to allow me to inject the aws transport 
  * I need to be clever about this in case I need to disable certain tests -- for example, due to the fact that AWS transport needs a connection_ack (or something like that)

**Questions:**
* Why are websocket tests makred as "asyncio" ?
* How does parameterization work?  I don't understand "server" as the parameter with "client_and_server" as the argument name.  I'm not sure how these functions are getting injected...
* Why does "server_no_ack" return nothing?
* How is "test_websocket_sending_invalid_data" asserting anything?

**Things to look into**
* The other websocket test files: test_websocket_online, test_websocket_query, test_websocket_subscription
* Whether I want to use parameterization to inject the new websocket transport or if I want to architect something more elegant

chadfurman:
Going to look into the other test_websocket_*.py files.   Ideally, I'd be able to just wrap them in a loop of sorts which would be able to run the same set of tests on different transports.  Or I'd be able to inject the transport as a parameter during parameterization (likely less impact) assuming I can understand how the pytest parameterization stuff is actually working -- I'll need to read the docs on that a bit better.

chadfurman:
**Learnings from reading test_websocket_online.py and the pytest docs**
* WSS and AWS URL format needs to match (Liskov substitution)
* Invalid data on a separate task can break parallel queries
* Indirect parameterization configures the fixture, not the test
* I should likely update fixtures and/or the parameterization for the tests to inject the AWS transport

**Answers to previous questions**
* Why are websocket tests makred as "asyncio" ? 
    * pytest-asyncio is a library that lets you use coroutines as tests / in your tests
* How does parameterization work? I don't understand "server" as the parameter with "client_and_server" as the argument name. I'm not sure how these functions are getting injected...
    * They're being injected into the fixtures that the tests depend on using indirect parameterization.  More here: https://docs.pytest.org/en/6.2.x/example/parametrize.html#indirect-parametrization
* Why does "server_no_ack" return nothing?
    * It's not a test, it's a fixture.  It's a fake server that doesn't ack.  It does not need to return anything.
* How is "test_websocket_sending_invalid_data" asserting anything?
    * I'm not sure that it is.  Instead, it's printing information.  This was likely used to test the code manually while it was being written rather than actually asserting functionality in the long-term.

**Outstanding questions**
* What are "online" tests?  What is "countries.trevorblades.com"?
* Should I extend the WebsocketServer fixture?
* Why do some tests use `server` and not `client_and_server`?  If I could standardize this, I could likely use indirect parameterization to inject the client I want to test.  
* If I am going to parameterize the client_and_server fixture, how would this look?  Currently, `client_and_server` is relying on the `server` fixture -- I'm thinking something like "parameterizing specific arguments" and having a "client" fixture?  
    * for reference https://docs.pytest.org/en/6.2.x/example/parametrize.html#apply-indirect-on-particular-arguments

Things to look into:
* Remaining test_websocket_*.py files: test_websocket_query.py, test_websocket_subscription.py
* Adding the client fixture and using client_and_server as the fixture in all tests (not the online tests), then parameterizing this
* Writing a test_awswebsocket_online.py test case that hooks up using actual creds to an actual endpoint and does actual stuff
    * How can I make an AWS CloudFormation template for running this test?  Where would I store it?

leszekhanusz:
>     * How is "test_websocket_sending_invalid_data" asserting anything?
>       
>       * I'm not sure that it is.  Instead, it's printing information.  This was likely used to test the code manually while it was being written rather than actually asserting functionality in the long-term.

This is a test which simulates the backend sending a `connection_error` when no queries are currently pending.
In that case, the exception cannot be returned and is silently ignored. The client closes the connection.

We simply verify here that no exceptions are returned in the user code. If any exception happens, then the test fails.

We could add some asserts here, to verify that the websocket connection is indeed closed and that the `close_exception` attribute is set to `TransportServerError`

>     * What are "online" tests?  What is "countries.trevorblades.com"?

"countries.trevorblades.com" is a dummy GraphQL backend to be used as an example. It is not affiliated with gql.
We use it for our examples and it was also used in the CI tests previously.

Because the tests using this backend can be flaky (the server could be down at any time), and to avoid overloading the server, I marked all those tests as "online" meaning it necessitates a working internet connection and the server working.

Now the automated tests done by GitHub Actions don't run the online tests, while still having 100% code coverage.

If you still want to run all the test, including the online tests, you can run `make all_tests`, which will run pytest with `--run-online` to not skip the online tests.

>     * Why do some tests use `server` and not `client_and_server`?  If I could standardize this, I could likely use indirect parameterization to inject the client I want to test.

client_and_server was created to reduce the copy/paste of the transport, client and session code to reduce the lines in a lot of tests. it will create a default session for us.

For some tests, we really need to instantiate the transport and the Client ourselves, and create the session ourselves, so we use `server` (for example `test_websocket_server_does_not_send_ack` is impossible with client_and_server, even if we could somehow parametrize the client inside the server fixture)

So please don't modify all the existing tests and instead use a new fixture for your tests :pleading_face:

chadfurman:
Understood, thank you

leszekhanusz:
> Do you care if I add another param to the existing websocket tests to re-use them and verify the AWS child-class passes Liskov Substitution?

Sorry, I don't really understand what you mean. How could the existing websockets tests work with your new class? It won't because it uses another protocol. So Liskov substitution would fail, but this is totally normal.

Tests are a domain where [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) is not useful. There are no problem to copy/paste and I don't see why you should modify the existing tests.

chadfurman:
So I tracked the issue to the gobal imports in my test cases.  I need to move the imports to inside my test cases which necessitates creating the fixtures.  Atleast I figured out why it's choking on my imports in the other tests, now.  I'll circle back to this another night.  Glad I figured it out, though.  And the linting error is something about a merge from master I think.  I'll fix all this later and make another push~

joseph-wortmann:
I'm back out of the hole I was in since September, if there is anything that I can do to help I am willing, @chadfurman

chadfurman:
> I'm back out of the hole I was in since September, if there is anything that I can do to help I am willing, @chadfurman

I'd love it if you could please drive this over the finish line, @joseph-wortmann 

My notes are in the PR thread.  Last comment is where I left off.  Was trying to increase test coverage and then was going to add documentation.

Also needs some solid AWS testing.  I was hoping the e2e tests would sort that out as I added them.  I.e. add the e2e tests for AWS and fix anything that causes them not to work.

leszekhanusz:
@joseph-wortmann are you working on this or should I try to complete this PR ? I've nearly finished with my other remaining gql tasks.

joseph-wortmann:
@leszekhanusz unfortunately I am not. As soon as I posted that I had some time life conspired to take it back from me. Have at it!

leszekhanusz:
It is now working with the API key version.
I've added some code examples from the default aws example. If you run the api_key_subscription.py in one terminal
and the api_key_mutation.py in another, you should see the "Hello World" message exchanged.

I am still trying to understand how to make it work with IAM.

chadfurman:
For IAM, if you setup a ~/.aws/config and .aws/credentials file with your IAM auth it should work.  There are likely errors but that's how you auth generally.  May need improvement, and I think the way our credentials object is written it can have creds passed in

leszekhanusz:
Trying to make it work with IAM:

* at first I had a `com.amazon.coral.service.http#HttpNotFoundException`
  * Fixed by the last commit setting the host correctly instead of the url
* Then I had a `com.amazonaws.deepdish.graphql.auth#UnauthorizedException`
  * Fixed by activating the IAM auth correctly with `amplify update api` followed by `amplify push`

Now I have `com.amazonaws.deepdish.graphql.auth#BadRequestException', 'message': "Credential should be scoped to correct service: 'appsync'`

Investigating...

leszekhanusz:
I can now connect using IAM and do the init but my first subscription query fails with:
`{"id":"1","type":"error","payload":{"errors":[{"errorType":"UnauthorizedException","message":"Permission denied"}]}}`

I suspect that somehow the IAM user is not allowed to execute this query in AWS.

leszekhanusz:
I don't really know what I am doing with the permissions. Here is what I did:

In AWS console:                                                                                                         
 - I have a single IAM user named **appsync_test**                                                                          
 - There is a role **MessageIAMRoleXXXXXX-XXXXXXXXXXXXXXXXXXXXXXXXXX-dev** that was created for me                          
                                                                                                                        
In this role, I modified the trust relationship to add my IAM user arn:                                                 
                                                                                                                        
```
{                                                                                                                       
  "Version": "2012-10-17",                                                                                              
  "Statement": [                                                                                                        
    {                                                                                                                   
      "Effect": "Allow",                                                                                                
      "Principal": {                                                                                                    
        "Service": "appsync.amazonaws.com",                                                                             
        "AWS": "arn:aws:iam::XXXXXXXXXXXX:user/appsync_test"                                                            
      },                                                                                                                
      "Action": "sts:AssumeRole"                                                                                        
    }                                                                                                                   
  ]                                                                                                                     
}           
```                                                                                                            
                                                                                                                        
On the console, I ran this command to assume the MessageIAMRole for my IAM user [to have a temporary token](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)
                                                                                                                        
`aws sts assume-role --role-arn arn:aws:iam::XXXXXXXXXXXX:role/MessageIAMRoleXXXXXX-XXXXXXXXXXXXXXXXXXXXXXXXXX-dev --role-session-name "RoleSession1" --profile appsync_test --duration-seconds 43200 > assume-role-output.txt`
                                                                                                                        
From the `assume-role-output.txt`, I created the environment variables **AWS_ACCESS_KEY_ID**, **AWS_SECRET_ACCESS_KEY** and **AWS_SESSION_TOKEN**
                                                                                                                        
My schema file backend/api/myrealtimeapp/schema.graphql is the default:                                                 
                                                                                                                        
```
# This "input" configures a global authorization rule to enable public access to                                        
# all models in this schema. Learn more about authorization rules here: https://docs.amplify.aws/cli/graphql/authorization-rules
input AMPLIFY { globalAuthRule: AuthRule = { allow: public } }                                                          
                                                                                                                        
type Message @model {                                                                                                   
    id: ID!                                                                                                             
    message: String!                                                                                                    
    createdAt: String                                                                                                   
} 
```

* What am I missing ?
* Should I replace **RoleSession1** in the above command by something else ?

chadfurman:
Interesting, so I would have assumed that appsync would make this a bit easier but then shame on me.

I'll have to dig into appsync IAM permissions a bit, but in the AWS console you should be able to make an IAM user with "full appsync permissions" and its own key/secret/id which "should work". Temporary user tokens and assume role stuff is definitely more secure but much harder to get right and for what is happening in this code base generally unnecessary (good to catch the permission error maybe or just let it bubble up)

leszekhanusz:
Note: I merged the two JWT classes into one as from our point of view it is the same thing.

Also I have something strange, the test `tests/test_appsync_auth.py::test_appsync_init_with_iam_auth_and_no_region` is somehow failing on my machine because it seems botocore finds my region, even if I rename my `.aws/config` file.

leszekhanusz:
For Reference my `Permission denied `error with IAM was fixed by adding the `AWSAppSyncInvokeFullAccess` policy to my IAM user.

Next, I add the following error:
`{"errorType":"Unauthorized","message":"Not Authorized to access onCreateMessage on type Message"}`

This one was solved by creating the `amplify/backend/api/<your-api-name>/custom-roles.json` file as described [here](https://docs.amplify.aws/cli/graphql/authorization-rules/#use-iam-authorization-within-the-appsync-console)

Now I can execute queries, mutations and subscriptions with IAM.

leszekhanusz:
The` appsync.py `file was split in 2 so that we could use the AppSync auth classes either with the websockets transport or with the aiohttp transport without needing the other dependency.

leszekhanusz:
@joseph-wortmann @chadfurman could you please check that the PR is ok for you as it is now? Do you have remarks?
I did not test the JWT auth. @joseph-wortmann could you please test it as it seems you have some jwts ?
I would like to merge this Friday.

chadfurman:
I've been following your commits and have no reason to object to anything here.  If there's functionality missing / needed / broken it will show during usage more than my review and so will lend itself to iteration.  I think this is fine as is.  Great work and thank you for taking it over.

PR #241, opened 2021-09-15
Features/filename upload

pkucmus:
Hi, I made this in dire need to specify the filename when uploading to a service I need to use. Wanted to add a test or two but had no idea how to wrap my head around those, I could use mocks to test this but was not sure you'd like that.


leszekhanusz (approved):

leszekhanusz:
Looks good. Thanks

PR #244, opened 2021-09-27
Adds upload_files functionality for request transport

DENKweit:
@leszekhanusz This PR requires a new package dependency: `requests_toolbelt` but I'm not sure where to include it. Could you add it for me? Thanks!

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -1,8 +1,11 @@
+import io
 import json
 import logging
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, Generator, Optional, Tuple, Type, Union
 
+import aiohttp
 import requests
+from requests_toolbelt.multipart.encoder import MultipartEncoder
 from graphql import DocumentNode, ExecutionResult, print_ast
 from requests.adapters import HTTPAdapter, Retry
 from requests.auth import AuthBase
@@ -10,6 +13,7 @@ from requests.cookies import RequestsCookieJar
 
 from gql.transport import Transport
 
+from ..utils import extract_files
 from .exceptions import (
     TransportAlreadyConnected,
     TransportClosed,
@@ -26,6 +30,10 @@ class RequestsHTTPTransport(Transport):
 
     The transport uses the requests library to send HTTP POST requests.
     """
+
+    file_classes: Tuple[Type[Any], ...] = (
+        io.IOBase,
+    )
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


leszekhanusz:
> @leszekhanusz This PR requires a new package dependency: `requests_toolbelt` but I'm not sure where to include it. Could you add it for me? Thanks!

I don't have access to a pc right now but you could add it to the `install_requests_requires` array in setup.py

DENKweit:
@leszekhanusz Is the decrease in coverage a problem? If yes, could you help me out here? Not sure what to change to get it back to 100%..

Also `make check` is failing at `flake8 gql tests scripts/gql-cli docs/code_examples` with:
```
flake8.exceptions.PluginRequestedUnknownParameters: "pycodestyle" requested unknown parameters causing 'FileProcessor' object has no attribute 'indent_size'
```
I used the conda environment which you are describing in `CONTRIBUTING.md`

leszekhanusz:
For the coverage, you could refactor:

```
            if self.headers is None:
                post_args["headers"] = {"Content-Type": data.content_type}
            else:
                post_args["headers"]["Content-Type"] = data.content_type
```

into:

```
            if post_args["headers"] is None:
                post_args["headers"] = {}

            post_args["headers"]["Content-Type"] = data.content_type
```

setup.py
```diff
@@ -38,6 +38,7 @@
 
 install_requests_requires = [
     "requests>=2.23,<3",
+    "requests_toolbelt>=0.9.1",
```
leszekhanusz, line 41: Please add `,<1` here so that a new major release of requests_toolbelt will not break gql.

leszekhanusz (changes requested):
LGTM 🚀
Just please add a maximum version for the requests_toolbelt dependency.

setup.py
DENKweit: done

leszekhanusz (approved):

leszekhanusz:
Thanks, I'll merge this PR next week.

maaft:
Hi! I noticed something strange with this PR.. The content-type header gets set correctly to "multipart/form-data" when you do an upload mutation, but when you execute another mutation/query on the same client afterwards, this header is **still present**. 

Looking at the code, I fail to grasp how this can be the case as `post_args` exists only in the local context and should not have any impact after the mutation was send.. @leszekhanusz  @DENKweit  do you have any idea how this can happen??

leszekhanusz:
Indeed, `post_args["headers"]` is set as a reference to` self.headers`

DENKweit:
Thank you for finding this! So, this is happening:
- `post_args["headers"]` is initialized to `self.headers`
- self.sesssion is initialized with that object during the next call to `self.session.request(...)`
- `self.session.request(...)` is internally updating `self.session.headers` therefore updating also `client.headers`

Writing a fix now.

leszekhanusz:
Thanks @DENKweit, @maaft !

PR #247, opened 2021-10-07
tansport aiohttp transport, single line change. ignore backend mimetype

mirkan1:
same trick didnt work for requests so only in aiohttps for now. waiting new version to be available on pypi because it is a dependancy, thank you :)


mirkan1:
@leszekhanusz

mirkan1:
AIOHTTPTransport should accept retries too

leszekhanusz:
a PR should only contain the relevant code change and tests.

Please don't upgrade the version and don't modify .gitignore

I would like to add a test also. I can do it if you want.

> AIOHTTPTransport should accept retries too

It should not accept retries. There are a million different ways to do retries (with jitter, with exponential backoff, only in some cases but not others, ...). If you need retries, then you can use [backoff](https://github.com/litl/backoff) (See [async advanced usage](https://gql.readthedocs.io/en/v3.0.0a6/advanced/async_advanced_usage.html) for an example)

mirkan1:
> a PR should only contain the relevant code change and tests.
> 
> Please don't upgrade the version and don't modify .gitignore
> 
> I would like to add a test also. I can do it if you want.
> 
> > AIOHTTPTransport should accept retries too
> 
> It should not accept retries. There are a million different ways to do retries (with jitter, with exponential backoff, only in some cases but not others, ...). If you need retries, then you can use [backoff](https://github.com/litl/backoff) (See [async advanced usage](https://gql.readthedocs.io/en/v3.0.0a6/advanced/async_advanced_usage.html) for an example)

I mean async calls by reties, sorry.
yeah sure let me delete some lines, version, and gitignore
write your own test if you want but fast please. Its been a while since I wrote about my project because of this issue.
also content_type=None doesnt work for requests library, gives error

PR #255, opened 2021-11-15
DSL serialize complex arguments to literals

### LINKED ISSUE #137: SequelizeJSON support
Hello!

Does GQL package (3.0.0a1) have the support of making queries with **DSL** for **SequelizeJSON type**? For example:
```graphql
{
  user_raw_search(where: {county: {eq: "Pulaski"}}) {
    user_id
  }
}
```
Where `where` argument is SequelizeJSON

Because in case of using in such way:
```python
dsl.query(
        dsl.Query.user_raw_search(where={"county": {"eq": "Pulaski"}}).select(
            dsl.User.user_id,
        ),
)['user_raw_search']
```
I got an error: _Cannot convert value to AST: {'county': {'eq': 'Pulaski'}}._

GraphQL schema:
```graphql
directive @cacheControl(
  maxAge: Int
  scope: CacheControlScope
) on FIELD_DEFINITION | OBJECT | INTERFACE
directive @specifiedBy(url: String!) on SCALAR
enum CacheControlScope {
  PUBLIC
  PRIVATE
}

type Query {
  user(id: ID!): User
  users(order: String, limit: Int, offset: Int): [User]
  user_search(where: UserQuery!, order: String, limit: Int, offset: Int): [User]
  user_raw_search(
    where: SequelizeJSON!
    order: String
    limit: Int
    offset: Int
  ): [User]
}

scalar SequelizeJSON

scalar Upload

type User {
  active: Boolean
  address: String
  address2: String
  address3: String
  city: String
  county: String
  dob: String
  email: String
  firstName: String
  gender: String
  id: ID
  lang: String
  lastName: String
  memberId: String
  phoneNumber: String
  sb360userId: String
  state: String
  zip: String
}

input UserQuery {
  active: Boolean
  address: String
  address2: String
  address3: String
  city: String
  county: String
  dob: String
  email: String
  firstName: String
  gender: String
  id: ID
  lang: String
  lastName: String
  memberId: String
  phoneNumber: String
  sb360userId: String
  state: String
  zip: String
}
```

leszekhanusz:
**NOTE: This PR depends on PR #253 and should be merged after it !**

This is a new feature for the DSL module which will attempt to serialize complex objects as literal (inside the query itself)
when provided as arguments of DSLFields.

It should fix #137

Note:
 - some scalars could have different serialization as variable values or as literals (it should be quite rare in practice). In that case, this method does not work.
 - there are some limitations, not all json values can be converted as literal inside the query (for example if the json value contain object with keys with spaces)
 - once the [graphql-js PR 3065](https://github.com/graphql/graphql-js/pull/3065) has been merged, and ported to graphql-core, then we could use it to allow custom serialization to literals for custom scalars

### DIFF
--- a/docs/modules/gql.rst
+++ b/docs/modules/gql.rst
@@ -21,3 +21,4 @@ Sub-Packages
    client
    transport
    dsl
+   utilities
--- a/docs/modules/utilities.rst
+++ b/docs/modules/utilities.rst
@@ -1,0 +1,6 @@
+gql.utilities
+=============
+
+.. currentmodule:: gql.utilities
+
+.. automodule:: gql.utilities
--- a/docs/usage/custom_scalars.rst
+++ b/docs/usage/custom_scalars.rst
@@ -1,0 +1,134 @@
+Custom Scalars
+==============
+
+Scalar types represent primitive values at the leaves of a query.
+
+GraphQL provides a number of built-in scalars (Int, Float, String, Boolean and ID), but a GraphQL backend
+can add additional custom scalars to its schema to better express values in their data model.
+
+For example, a schema can define the Datetime scalar to represent an ISO-8601 encoded date.
+
+The schema will then only contain:
+
+.. code-block:: python
+
+    scalar Datetime
+
+When custom scalars are sent to the backend (as inputs) or from the backend (as outputs),
+their values need to be serialized to be composed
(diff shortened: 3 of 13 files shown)


lenoirmind:
Verstuurd vanaf mijn iPad

PR #256, opened 2021-11-19
Parse custom scalar and enums in outputs

leszekhanusz:
### Adding parse_results functionality
If you add `parse_results=True` to the Client init,
or if you add `parse_result=True` to the execute or subscribe methods,
then gql will unserialize the result received from the backend with the scalars defined in the schema.

Note: because [graphql-core PR 138](https://github.com/graphql-python/graphql-core/pull/138) is not yet in a stable release, you might have issues with enums (parsed to `None`) if you get your schema from introspection.

### New utilities method to update the schema

- `update_schema_enum` to update an Enum in the schema
- `update_schema_scalar` to update a Scalar in the schema


leszekhanusz:
Hmm, there are a bunch of problems with this PR... working on a fix.

Issue #206, opened 2021-05-08
gql-cli pagination


abitrolly:
Looks like https://gql.readthedocs.io/en/latest/gql-cli/intro.html doesn't support pagination, which is necessary to get all results from API calls like GitLab https://docs.gitlab.com/ee/api/graphql/getting_started.html#pagination in one go.

Are there any plans to add it?

inesusvet:
Could you please provide an example of the query, variables and the gql usage for this API call

abitrolly:
That should work for a query. The project https://gitlab.com/gitlab-org/growth/product/-/issues contains about 500 issues and it should not run over API requests limits if executed several times.
```
query {
  project(fullPath: "gitlab-org/growth/product") {
    name
    issues {
      edges {
        node {
          title
        }
      }
      pageInfo {
        endCursor
        hasNextPage
      }
    }
  }
}
```
Variables:
- the name of the project
- command line flag `--pages` to tell that pagination should be used
  - --pages=all - all pages

Not sure if the number of pages should start with 0 or 1. :D

abitrolly:
Workspace setup.
```
git clone https://github.com/graphql-python/gql
cd qgl
pipenv install -e .
```

leszekhanusz:
Pagination can be different for each type of schema, so it will not be implemented in gql-cli which should be generic for all types of schemas.

But you can modify your query to add variables and pass variables from the command line. Like this:

```
echo '
query getIssues ($first: Int, $after: String) {
  project(fullPath: "gitlab-org/growth/product") {
    name
    issues (first: $first, after: $after) {
      nodes {
        title
      }
      pageInfo {
        endCursor
        hasNextPage
      }
    }
  }
}
' > issues.gql
cat issues.gql | gql-cli https://gitlab.com/api/graphql --variables first:2 after:eyJpZCI6IjgwNDc5NDc3IiwiY3JlYXRlZF9hdCI6IjIwMjEtMDMtMDggMTQ6NTI6NDUuMjY2ODQ5MDAwIFVUQyJ9
```

abitrolly:
@leszekhanusz where do you get that `after:eyJpZCI6IjgwNDc5NDc3IiwiY3JlYXRlZF9hdCI6IjIwMjEtMDMtMDggMTQ6NTI6NDUuMjY2ODQ5MDAwIFVUQyJ9` from?

abitrolly:
[...]

PR #269, opened 2021-11-30
Document absence of pagination support in `gql-cli`

abitrolly:
Ref. #206


leszekhanusz:
That is false, pagination is supported by gql, See [my comment](https://github.com/graphql-python/gql/issues/206#issuecomment-846452711) for an example on how to do it.

abitrolly:
@leszekhanusz #271 PTAL.

Issue #252, opened 2021-10-31
How to query with multiple levels of nested connections?


svcosmghtest:
Hi guys, how to use this library to query with multiple levels of nested connections?

I tried to get the org data in the following way:
```python
      client = get_gh_graphql_api_client()
      active_vul_alerts = set()
      with client as graphql_client_session:
          ds = DSLSchema(graphql_client_session.client.schema)
          page_query_params = {"first": settings.GH_GRAPHQL_PAGE_SIZE}
          has_next_page = True
          while has_next_page:
              query = dsl_gql(
                  DSLQuery(
                      ds.Query.organization(login=org_login).select(
                          ds.Organization.repositories(**page_query_params).select(
                              ds.RepositoryConnection.pageInfo.select(ds.PageInfo.hasNextPage, ds.PageInfo.endCursor,),
                              ds.RepositoryConnection.nodes.select(
                                  ds.Repository.vulnerabilityAlerts.select(
                                      ds.RepositoryVulnerabilityAlertConnection.nodes.select(
                                          ds.RepositoryVulnerabilityAlert.dismissedAt,
                                          ...
                                      ),
                                  ),
                                  ds.Repository.databaseId,
                                  ds.Repository.name,
                              ),
                          )
                      )
                  )
              )
              result = graphql_client_session.execute(query)
              organization = result["organization"]
              repository_connection = organization["repositories"]
              page_info = repository_connection["pageInfo"]
              end_cursor = page_info["endCursor"]
              if not end_cursor:
                  # This is the first page and there is no data in this page
                  break
              page_query_params["after"] = end_cursor
              has_next_page = page_info.get("hasNextPage")
              ...
```

the following error will raise:
```
TransportQueryError:
{
  'type': 'MISSING_PAGINATION_BOUNDARIES',
  'path': ['organization', 'repositories', 'nodes', 0, 'vulnerabilityAlerts'],
  'locations': [{'line': 19, 'column': 9}],
  'message': 'You must provide a `first` or `last` value to properly paginate the `vulnerabilityAlerts` connection.'
}
```

I also want to know if this is the best way to do the pagination or there is some advanced way to do it? Thanks!

leszekhanusz:
This is not a gql issue.
You need to provide a `first` argument for vulnerabilityAlerts too. See [Resource limitations](https://docs.github.com/en/graphql/overview/resource-limitations) in GitHub API docs.

svcosmghtest:
Thank you for your response, Hanusz.
[...]

PR #271, opened 2021-11-30
Document that `gcl-cli` will not implement pagination

abitrolly:
Ref. #269.

### DIFF
--- a/docs/gql-cli/intro.rst
+++ b/docs/gql-cli/intro.rst
@@ -8,6 +8,8 @@ GraphQL queries directly from the terminal.
 
 This script supports http(s) or websockets protocols.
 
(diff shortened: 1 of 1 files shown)


abitrolly:
@leszekhanusz any comment why you closed this without merging?

It is needed for people who are looking for a CLI tool to fetch paginated results without additional scripting.

leszekhanusz:
I've already explained myself in your previous PR #269 
You don't understand that it is not possible to implement it a universal way to fetch pages because each backend can implement pagination how it likes.
And what should we do for nested pagination as in #252 ? RIGHT that's not possible.

You can easily implement whatever pagination support you want for whatever backend you want with a little python script.
I won't do your homework for you.

Stop spamming this repository or I will block you.

If your not happy with this, feel free to fork this repo.

PR #274, opened 2021-12-06
⬆️ Expand aiohttp version range

connorbrinton:
A new version of aiohttp has been released, with [an important fix preventing header injection](https://github.com/aio-libs/aiohttp/issues/4818). I looked through the aiohttp changelog and wasn't able to find any breaking changes, so I believe it should be safe to use with gql.

All of the offline tests passed locally with aiohttp 3.8.1 installed 🙂 

Thanks for your work on gql! 😄

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ dev_requires = [
 ] + tests_requires
 
 install_aiohttp_requires = [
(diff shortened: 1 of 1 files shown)


leszekhanusz (approved):

leszekhanusz:
Thanks for your PR

PR #275, opened 2021-12-06
Enable mypy to discover type hints as specified in PEP 561

leszekhanusz:
This should enable mypy to discover type hints for the users of this library.


srittau (approved):
Thank you, LGTM!

Issue #278, opened 2021-12-09
Validation error using '@include' directive if schema comes from introspection


SuperBo:
When I try to use "@include" directive in my string query, I got this error:
`graphql.error.graphql_error.GraphQLError: Unknown directive '@include'.`

I think "@include" and "@skip" are in standards of GraphQL https://graphql.org/learn/queries/#directives

Version in my environment

```
gql[requests]==3.0.0b1
graphql-core==3.1.6; python_version >= '3.6' and python_version < '4'
```

**GraphQL Query**

```graphql
query fetchPool($with_swap: Boolean!) {
  pool(id: "0x83abecf7204d5afc1bea5df734f085f2535a9976") {
    id
    createdAtTimestamp
    swaps (first: 2) @include(if: $with_swap ) {
      id
      timestamp
    }
  }
}
```

**Full stack trace:**

> Traceback (most recent call last):
>   File "test.py", line 27, in <module>
>     result = client.execute(q, variable_values= {'with_swap': True})
>   File "/Users/ynguyen/.local/share/virtualenvs/activelp_v1_prod-cEysbnrO/lib/python3.8/site-packages/gql/client.py", line 186, in execute
>     return self.execute_sync(document, *args, **kwargs)
>   File "/Users/ynguyen/.local/share/virtualenvs/activelp_v1_prod-cEysbnrO/lib/python3.8/site-packages/gql/client.py", line 134, in execute_sync
>     return session.execute(document, *args, **kwargs)
>   File "/Users/ynguyen/.local/share/virtualenvs/activelp_v1_prod-cEysbnrO/lib/python3.8/site-packages/gql/client.py", line 389, in execute
>     result = self._execute(
>   File "/Users/ynguyen/.local/share/virtualenvs/activelp_v1_prod-cEysbnrO/lib/python3.8/site-packages/gql/client.py", line 328, in _execute
>     self.client.validate(document)
>   File "/Users/ynguyen/.local/share/virtualenvs/activelp_v1_prod-cEysbnrO/lib/python3.8/site-packages/gql/client.py", line 129, in validate
>     raise validation_errors[0]
> graphql.error.graphql_error.GraphQLError: Unknown directive '@include'.
> 
> GraphQL request:6:22
> 5 |     createdAtTimestamp
> 6 |     swaps (first: 2) @include(if: $with_swap ) {
>   |                      ^
> 7 |       id

leszekhanusz:
Hi,

[...]

PR #279, opened 2021-12-09
Add skip and include directive in introspection schema

leszekhanusz:
Workaround for issue #278 until [graphql-js#3419](https://github.com/graphql/graphql-js/issues/3419) has been fixed.

Always adding `skip` and `include` default directives in schema built from introspection.


Cito:
Right, that makes sense.

PR #283, opened 2021-12-13
Bump graphql-core to 3.2.0

leszekhanusz:


leszekhanusz:
@cito
Should I release the stable 3.0.0 version of gql with graphql-core 3.1.x or should I wait for graphql-core 3.2 stable and merge this PR? What do you think?

Cito:
If everything is working fine with the 3.2 rc3 version, then I think it's ok - 3,2 final will be released soon and will not have breaking changes.

Cito:
Sorry, there was a 3.2 rc4 version, and contrary to what I promised above, it contained one breaking change. To make up for this, I have already updated this PR with the necessary fixes. And this should really be the last breaking change before 3.2 final.

leszekhanusz:
No worries, keep up the good work!

ddelange:
Does this PR drop support for `graphql-core==3.1.*`? Or could `gql` theoretically still allow `3.1.*` in `install_requires`? Doing so would solve dephell ref https://github.com/graphql-python/graphene/issues/1406

PR #285, opened 2021-12-17
Update type hint to align with docs and usage.

joaufi:
### Description

Updated the type hint for the `verify` kwarg of `RequestsHTTPTransport` to align with both the class documentation and intended usage of `verify`.

### Testing

All tests pass. I'm happy to write a test that verifies both a `bool` and a `str` can be provided as an arg to `verify` but that seemed a bit overkill to me as MyPy should cover such usages in application code.


leszekhanusz (approved):

leszekhanusz:
Thanks for your PR :+1:

PR #286, opened 2021-12-24
Backport to graphql-core-legacy (i.e. 2.x)

zdzislaw-s:
Hi There,

Within our company, we've been successfully using `gql` mainly as a client for handling subscriptions. Our setup includes `django-channels` and we couldn't get subscription to work with version 3 of the graphql modules (i.e. `graphql-core`, `gql`, `graphene`). For that reason we experimented with back-porting `gql` to use `graphql-core-legacy` and that worked for us. We hope that the changes here could be of use to some other folks out there.

This PR presents changes that we applied to the `gql` code to get subscriptions working with `graphql-core` v2+.

**Please note that it is not intention of this PR to be merged to master.**


leszekhanusz:
Hi, that's an interesting approach but unfortunately the 2.x version is not maintained anymore so I have to close this PR.

zdzislaw-s:
I understand that you don't maintain version 2.x, still if you accept the changes and make a branch in the repo where these changes could leave, people out would find it easier (in case they would like to work with _subscriptions_ and with `graphql-core-legacy`) to get code that would meet their needs. Otherwise, the only trace that it is possible to get subscriptions working with `graphql-core-legacy` is this PR and this has rather low visibility.

Anyway, it's your call after all, so _que sera sera_ :wink:

All the best in the New Year 2022 and great thanks for your `gql`!

leszekhanusz:
@zdzislaw-s 

I want to take responsibility for all the code which is merged in this repository.
If I put the code in another branch, then we could get issues to fix some problems on it and I don't want to work on obsolete code. I hope you understand.

Happy new year for you too!

PR #287, opened 2022-01-04
🏷 Add type overloads for `get_execution_result`

connorbrinton:
Methods like `execute` were previously annotated with a `Union` type, indicating all of the types of return values that the function could be returned. However, a more specific return type can often be inferred statically based on whether or not the `get_execution_result` keyword argument is passed to the method.

These changes add function type overloads to methods accepting the `get_execution_result` keyword, which allows Mypy to infer a more specific return type based on the passed value of `get_execution_result`, if any.


gql/client.py
```diff
@@ -525,7 +554,35 @@ async def _subscribe(
...
+        variable_values: Optional[Dict[str, Any]] = ...,
+        operation_name: Optional[str] = ...,
+        serialize_variables: Optional[bool] = ...,
+        parse_result: Optional[bool] = ...,
+        get_execution_result: Literal[True],
+        **kwargs,
+    ) -> AsyncGenerator[ExecutionResult, None]:
+        ...
+
+    async def subscribe(  # type: ignore[misc]
```
connorbrinton, line 585: This `# type: ignore` directive is required because Mypy is unable to infer that `subscribe` will only ever yield a single type for a particular instantiation of the generator. There's probably a complex way to make this work properly by moving out some functionality into another overloaded function, but I worry that it might overcomplicate the code.

gql/compat.py
```diff
@@ -0,0 +1,10 @@
+"""Compatibility helpers to smooth over differences between Python versions."""
```
connorbrinton, line 1: I'm open to changes to this filename, or combining it with an existing file 🙂

gql/client.py
leszekhanusz: It was solved by removing the async keyword in the overloaded definitions.

gql/compat.py
leszekhanusz: I prefer to keep it inside the client.py for now. We can still move it to another file if we need to use it elsewhere later.

leszekhanusz:
Thanks for your PR!

Note: tests are flaky on windows for unrelated reasons. I'll verify this and probably disable this test for windows.

leszekhanusz (approved):

connorbrinton:
@leszekhanusz You bet, thanks for tidying it up and maintaining gql! 😄

PR #293, opened 2022-02-05
Saving http response headers reference in transports

sondale-git:
The purpose of this Pull Request is to add a request response header attribute. Accessing response header can be useful for retrieving rate limiting information from some APIs.


leszekhanusz:
Thanks for your PR! I added the reference to the other transports as well and added some tests and documentation.

leszekhanusz (approved):

sondale-git:
Thank you for your reply and contributions! :smile:

PR #297, opened 2022-02-18
Close transport when fetching the schema failed

joricht:
Fetching the schema from the transport might fail with an exception. This Commit
ensures that the transport is closed in such a case and the client object can be
used to open a new session.

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -275,6 +275,9 @@ class Client:
             if self.fetch_schema_from_transport and not self.schema:
                 await self.session.fetch_schema()
         except Exception:
+            # we don't know what type of exception is thrown here because it
+            # depends on the underlying transport; we just make sure that the
+            # transport is closed and re-raise the exception
             await self.transport.close()
             raise
 
@@ -301,6 +304,9 @@ class Client:
             if self.fetch_schema_from_transport and not self.schema:
                 self.session.fetch_schema()
         except Exception:
+            # we don't know what type of exception is thrown here because it
+            # depends on the underlying transport; we just make sure that the
+            # transport is closed and re-raise the exception
             self.transport.close()
             raise
 
--- a/gql/client.py
+++ b/gql/client.py
@@ -271,8 +271,12 @@ class Client:
             self.session = AsyncClientSession(client=self)
 
         # Get schema from transport if needed
-        if self.fetch_schema_from_transport and not self.schema:
-            await self.session.fetch_schema()
+        try:
+            if self.fetch_schema_from_transport and not self.schema:
+                await self.session.fetch_schema()
+        except Exception:
+            await self.transport.close()
+            raise
 
         return self.session
 
@@ -293,8 +297,12 @@ class Client:
             self.session = SyncClientSession(client=self)
 
         # Get schema from transport if needed
-        if self.fetch_schema_from_transport and not self.schema:
-            self.session.fetch_schema()
+        try:
+            if self.fetch_schema_from_transport and not self.schema:
+                self.session.fetch_schema()
+        except Exception:
+            await self.transport.close()
+            raise
 
         return self.session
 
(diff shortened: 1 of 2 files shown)


gql/client.py
```diff
@@ -293,8 +297,12 @@ def __enter__(self):
             self.session = SyncClientSession(client=self)
 
         # Get schema from transport if needed
-        if self.fetch_schema_from_transport and not self.schema:
-            self.session.fetch_schema()
+        try:
+            if self.fetch_schema_from_transport and not self.schema:
+                self.session.fetch_schema()
+        except Exception:
+            await self.transport.close()
```
leszekhanusz, line 304: Should be without the `await` of course

gql/client.py
```diff
@@ -271,8 +271,12 @@ async def __aenter__(self):
             self.session = AsyncClientSession(client=self)
 
         # Get schema from transport if needed
-        if self.fetch_schema_from_transport and not self.schema:
-            await self.session.fetch_schema()
+        try:
+            if self.fetch_schema_from_transport and not self.schema:
+                await self.session.fetch_schema()
+        except Exception:
```
leszekhanusz, line 277: Please specify `(TransportError, GraphQLError)` instead of `Exception`

gql/client.py
```diff
@@ -293,8 +297,12 @@ def __enter__(self):
             self.session = SyncClientSession(client=self)
 
         # Get schema from transport if needed
-        if self.fetch_schema_from_transport and not self.schema:
-            self.session.fetch_schema()
+        try:
+            if self.fetch_schema_from_transport and not self.schema:
+                self.session.fetch_schema()
+        except Exception:
```
leszekhanusz, line 303: Please specify `(TransportError, GraphQLError)` instead of `Exception`

leszekhanusz (changes requested):
Thanks for your PR!

Could you please also add two tests for sync and async ?

gql/client.py
joricht: What a stupid copy and paste error :see_no_evil: sorry.

gql/client.py
joricht: Thanks for the suggestion. I actually wasn't sure what kind of exceptions could be thrown since it depends on the transport.

After thinking about it: In my example I used the AIOHTTPTransport, which throws a ClientError which inherits directly from Exception. This wouldn't be caught with your suggestion.

joricht:
@leszekhanusz Thank you for the quick reply and review. I am on holiday over the weekend and will address everything on Monday.

gql/client.py
leszekhanusz: You're perfectly right. I thought about fixing this kind of leaky abstractions by trying to catch all the exceptions of the transports and reraising them as TransportErrors but it can get tricky...

joricht: @leszekhanusz I checked what kind of error is thrown and it depends on the transport. `HttpRequestTransport` throws an `ConnectionError` from the underlying urllib3 and the `AIOHttpTransport` throws a custom `ClientError`. All of those inherit from the base exception. Therefore, I see no way to narrow down the exception here.

Does it matter any way? We want to make sure that the transport is closed on any error. And then we re-raise the exception.

joricht:
@leszekhanusz I added two tests and a comment why I catch the broad exception. Can you have a look at it again. Thank you!

leszekhanusz:
Thanks. Please read the [CONTRIBUTING.md](https://github.com/graphql-python/gql/blob/master/CONTRIBUTING.md) file to see how to fix linter errors (running `make check`)

joricht:
@leszekhanusz Thanks for being so patient. Now all tests should pass and the coding guidelines are also respected.

leszekhanusz (approved):

leszekhanusz:
LGTM :rocket:

PR #311, opened 2022-03-27
replace use of `scripts` with `entry_points`

bollwyvl:
## References

- fixes #310

## Changes
- [x] moves `scripts/gql-cli` to a `gql_cli` function in `gql/cli.py`
  - [x] actually delete `scripts/gql-cli`
- [x] update `setup.py`
- [x] update `MANIFEST.in`
- [x] update `Makefile`

Issue #310, opened 2022-03-25
Replace use of `scripts` with cross-platform `entry_points`


bollwyvl:
## Problem 

In looking into #309, I see that `scripts` are being used. These are troublesome for cross-platform downstreams, as they make otherwise-well-behaved distributions less portable.

## Proposal

Migrate from `scripts` to `entry_points`, where:

- the contents of `scripts/gql-cli` would be move to `gql/cli.py`
  - as `main` is already named there, might have a different name, e.g. `entry_point_main`
- `setup.py` would be updated, a la:

```diff
--- setup.py	2022-03-25 11:06:08.397539455 -0500
+++ setup.py	2022-03-25 11:08:23.321126291 -0500
@@ -8,7 +8,7 @@
 ]
 
 scripts = [
-    "scripts/gql-cli",
+    "gql-cli=gli.cli:main"
 ]
 
 tests_requires = [
@@ -106,5 +106,5 @@
     include_package_data=True,
     zip_safe=False,
     platforms="any",
-    scripts=scripts,
+    entry_points={"console_scripts": scripts}
 )

```

This would also allow the noarch python wheels to be created, which would be helpful for performance and security without increasing the maintenance/build burden much.

bollwyvl:
Opened #311

PR #312, opened 2022-03-27
build wheel

bollwyvl:
## References

- enabled by #311 (as old-style `scripts` decrease portability of built wheels)
- finishes up #310

## Changes

- [x] also builds wheel


leszekhanusz (approved):

bollwyvl:
Great! Now to track down more upstreams!

PR #328, opened 2022-05-16
Check for errors during client.fetch_schema()

luketaverne:
Hi everyone!

We (@wang2bo2 and I) came across a bug when our JWT tokens expired. 

## Current behavior / steps to reproduce
- Use an expired JWT token
- The server returns an error for the expired token, but the error is not checked client side (see PR changes). 
- The result is the error being added added to the client as the data/schema. 
- This causes another error to be raised, which isn't immediately obvious that it's due to an expired JWT. See stacktrace below.

```
... [project trace omitted] ...
File "/root/.cache/pypoetry/virtualenvs/project-xS3fZVNL-py3.8/lib/python3.8/site-packages/gql/client.py", line 617, in __enter__
         self.session.fetch_schema()
File "/root/.cache/pypoetry/virtualenvs/project-xS3fZVNL-py3.8/lib/python3.8/site-packages/gql/client.py", line 806, in fetch_schema
         self.client.schema = build_client_schema(self.client.introspection)
File "/root/.cache/pypoetry/virtualenvs/project-xS3fZVNL-py3.8/lib/python3.8/site-packages/gql/utilities/build_client_schema.py", line 79, in build_client_schema
         raise TypeError(
TypeError: Invalid or incomplete introspection result. Ensure that you are passing the 'data' attribute of an introspection response and no 'errors' were returned alongside: None.
```

## Expected behavior
- The error from the server should be raised inside the client sdk.

## Fix
The error handling used elsewhere in the library was copied over to the sync and async `fetch_schema` functions.

Thank you all for maintaining this useful SDK 🙂

### DIFF
--- a/docs/advanced/error_handling.rst
+++ b/docs/advanced/error_handling.rst
@@ -41,6 +41,11 @@ Here are the possible Transport Errors:
   The message of the exception contains the first error returned by the backend.
   All the errors messages are available in the exception :code:`errors` attribute.
 
+  If the error message begins with :code:`Error while fetching schema:`, it means
+  that gql was not able to get the schema from the backend.
+  If you don't need the schema, you can try to create the client with
+  :code:`fetch_schema_from_transport=False`
+
 - :class:`TransportClosed <gql.transport.exceptions.TransportClosed>`:
   This exception is generated when the client is trying to use the transport
   while the transport was previously closed.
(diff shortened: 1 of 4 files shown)


leszekhanusz:
Thanks for your PR!
I modified the message so that we could see directly that the error comes from fetching the schema and added some tests.

leszekhanusz (approved):

leszekhanusz:
Fixed in [v3.3.0](https://github.com/graphql-python/gql/releases/tag/v3.3.0)

Issue #347, opened 2022-07-28
Setting execution timeout with gql-cli


gonzalezzfelipe:
**Describe the bug**

There is a default `execution_timeout` on the client, which can't be modified on the CLI. This means that any request that lasts more than 10 seconds that is done using the `gql-cli` will fail, and there's nothing the user can do about it. It is possible to add the possiblity to modify this timeout using the CLI arguments, but why have the default timeout at all? I don't see any gain imposing a timeout on the users.

Refferences in the code:

* Default execution timeout in client: https://github.com/graphql-python/gql/blob/master/gql/client.py#L78
* Usage of the client on the CLI: https://github.com/graphql-python/gql/blob/master/gql/cli.py#L369-L371

**System info (please complete the following information):**
 - OS: MacOs
 - Python version: 3.10.4
 - gql version: 3.4.0
 - graphql-core version: 3.2.1

leszekhanusz:
Hmm it seems we did the same PR at the same time...

Adding arguments to gql-cli can be done.

Removing the timeout is not an option and increasing the default would be a breaking change and I don't see the point if it is configurable.

PR #349, opened 2022-07-28
Add execute-timeout argument for gql-cli

leszekhanusz:
See issue #347

### DIFF
--- a/gql/cli.py
+++ b/gql/cli.py
@@ -104,6 +104,13 @@ def get_parser(with_examples: bool = False) -> ArgumentParser:
         dest="print_schema",
     )
     parser.add_argument(
+        "--execute-timeout",
+        help="set the execute_timeout argument of the Client (default: 10)",
+        type=int,
+        default=10,
+        dest="execute_timeout",
+    )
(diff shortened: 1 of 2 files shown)


gql/cli.py
```diff
@@ -103,6 +103,13 @@ def get_parser(with_examples: bool = False) -> ArgumentParser:
         action="store_true",
         dest="print_schema",
     )
+    parser.add_argument(
+        "--execute-timeout",
+        help="set the execute_timeout argument of the Client (default: 10)",
+        type=int,
+        default=None,
```
gonzalezzfelipe, line 110: Shouldn't `default=None` be changed to `default=10`? If not the default would be no timeout. Also, should something like a `-1` be considered as `None`? That would give the user the possibility to set no timeout

leszekhanusz: You're right, I've changed back the default to 10.

To disable the timeout, I propose to allow `--execute-timeout=none` as it is done in my last commit.

gonzalezzfelipe (approved):
LGTM 👍

PR #350, opened 2022-08-04
Make `AsyncTransport` and `Transport` proper abstract base classes

willfrey:
The `@abc.abstractmethod` decorator requires that the class’s metaclass be `abc.ABCMeta` or derived from it. Usually this is satisfied by inherit from `abc.ABC`.

See: https://docs.python.org/3/library/abc.html#abc.abstractmethod


leszekhanusz:
Thanks, Well spotted!
Could you please add another commit here with the same modification for the `Transport` class?

leszekhanusz (approved):

PR #406, opened 2023-04-20
chore: Add assert message

levsa:


leszekhanusz:
Could you please give an explanation why you think that message in the assert is necessary?
In which case does the assert fails, do you have a small example?

levsa:
> Could you please give an explanation why you think that message in the assert is necessary? In which case does the assert fails, do you have a small example?

It helped me find which of the types I had added that caused the assertion to fail, and locate the problem. Here is a simple example:

```
import graphql
from gql.dsl import DSLSchema


schema_str = """
type FloatValue {
    floatValue: Float!
}

type IntValue {
    intValue: Int!
}

union Value = FloatValue | IntValue

type Entry {
    name: String!
    value: Value
}

type Query {
    values: [Entry!]!
}

"""
schema = graphql.build_ast_schema(graphql.parse(schema_str))
ds = DSLSchema(schema)
print(ds.Value)
```

Here, it is obvious that `Value` is accessed in `ds`, but when we have more complex code that accesses types in more automatic ways, it is hard to find the type that caused the assertion fail. Especially if I'm new to GraphQL and didn't understand that it was a `union` causing the failure.

leszekhanusz:
Thanks for your PR.
I think that it's best to change that `assert` into an `AttributeError`.
See the PR #409 

The error now should be more clear:

```
  File "test.py", line 28, in <module>
    print(ds.Value)
  File "/mnt/tera/git-repos/leszekhanusz/gql/gql/dsl.py", line 299, in __getattr__
    raise AttributeError(
AttributeError: Type "Value (<GraphQLUnionType 'Value'>)" is not valid as an attribute of DSLSchema. Only Object types or Interface types are accepted.
```

PR #407, opened 2023-04-21
Fix typo response_headers in httpx transport

cybniv:
Should be called `response_headers` for it to work :)

Thanks for this great piece of software ! ❤️


leszekhanusz (approved):
Well spotted, thanks.

PR #410, opened 2023-04-26
Python 3.11 support

leszekhanusz:


joeltorby:
Can someone confirm wether GQL works for python 3.11 or not? According to this issue it looks like it works, but according to documentation it looks like it doesn't. 
image

leszekhanusz:
I can confirm that the stable 3.4.1 gql version works with Python 3.11

That image from the README is autogenerated from the classifiers in the `setup.py` file from the last stable version (3.4.1)

We have some difficulties to generate a new stable version for now (See #427 for an explanation)

PR #421, opened 2023-07-12
Make retry backoff and status codes customizable

mmmeeedddsss:
RequestsHTTPTransport does not allow customization of urllib3's retry method. 

In this PR, I added two new parameters to RequestsHTTPTransport for supporting most of the possible customization cases when the RequestsHTTPTransport is used with retries: retry_backoff_factor & retry_status_forcelist. ([See urllib3's doc for Retry](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry))

Two notes on compatibility:
- I wanted to make this change backwards compatible as much as it could be. For the clients of gql that use `RequestsHTTPTransport` without keyed args, this change is backward incompatible since it changes the order of argument `method `, but I can refactor to move two new parameters to the place after `method`. In that case, the argument `retries ` and the new two arguments related to retries will be separated, which is not a huge deal.
- I believe most elegant solution for the customizability here would accept either an HTTPAdapter object or a Retry object from outside, but I didn't want to break the compatibility in such a huge way.

### DIFF
--- a/gql/transport/requests.py
+++ b/gql/transport/requests.py
@@ -1,7 +1,7 @@
 import io
 import json
 import logging
-from typing import Any, Dict, Optional, Tuple, Type, Union
+from typing import Any, Collection, Dict, Optional, Tuple, Type, Union
 
 import requests
 from graphql import DocumentNode, ExecutionResult, print_ast
@@ -31,6 +31,7 @@ class RequestsHTTPTransport(Transport):
     """
 
     file_classes: Tuple[Type[Any], ...] = (io.IOBase,)
+    _default_retry_codes = (429, 500, 502, 503, 504)
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


leszekhanusz:
Thanks for the PR!
You should be able to fix the lint issues by running `make check` as described in the [CONTRIBUTING.md](https://github.com/graphql-python/gql/blob/master/CONTRIBUTING.md) file.

mmmeeedddsss:
@leszekhanusz Now it should be okay, could you re-run the tests?

leszekhanusz:
>     * I wanted to make this change backwards compatible as much as it could be. For the clients of gql that use `RequestsHTTPTransport` without keyed args, this change is backward incompatible since it changes the order of argument `method `, but I can refactor to move two new parameters to the place after `method`. In that case, the argument `retries ` and the new two arguments related to retries will be separated, which is not a huge deal.

Yes, If you could please move the new arguments to the end that would be best. 
The probability of a breaking chance seems quite low in this case but you never know. 

>     * I believe most elegant solution for the customizability here would accept either an HTTPAdapter object or a Retry object from outside, but I didn't want to break the compatibility in such a huge way.

Good, that's fine like this.

mmmeeedddsss:
@leszekhanusz Made the change, thank you for the guidance. It should be fully backward compatible now

leszekhanusz (approved):

PR #425, opened 2023-07-25
Adjust aiohttp pin

cdce8p:
Update `aiohttp` pin to allow versions other then `3.8.x` with `<4`.

**Background**: It looks like aiohttp will add support for Python `3.12` only in `3.9.0`. By pinning `<3.9` this packages unnecessarily constraints what versions downstream packages or apps can install. With semantic versioning everything up to `<4.0` should be backwards compatible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ dev_requires = [
 ] + tests_requires
 
 install_aiohttp_requires = [
-    "aiohttp>=3.8.0,<3.9.0",
+    "aiohttp>=3.8.0,<4",
 ]
 
 install_requests_requires = [


leszekhanusz:
The 3.9 aiohttp version is not released yet, but they sure are [supposed to follow semver](https://docs.aiohttp.org/en/v3.8.5/faq.html#what-is-the-api-stability-and-deprecation-policy) so we might allow it.

But why did you want this change now before the version is available? Why not wait that the 3.9 version is available and verify that all the tests are working before doing this change?

cdce8p:
> The 3.9 aiohttp version is not released yet, but they sure are [supposed to follow semver](https://docs.aiohttp.org/en/v3.8.5/faq.html#what-is-the-api-stability-and-deprecation-policy) so we might allow it.
> 
> But why did you want this change now before the version is available? Why not wait that the 3.9 version is available and verify that all the tests are working before doing this change?

I've already starting testing Python 3.12 for [Home Assistant](https://github.com/home-assistant/core). Since we have a lot (>880) direct dependencies, it's quite common that a new Python version doesn't work initially. The earlier we can fix potential issues, the easier it will be to fully support it later on once 3.12 is released. 

Just for reference, this is the aiohttp PR https://github.com/aio-libs/aiohttp/pull/7400 which added build support for Python `3.12`. It was only tagged to be backported to  `3.9` not `3.8`. That's why I already checked which of our dependencies aren't compatible.

leszekhanusz (approved):

leszekhanusz:
This is now available in [pre-release 3.5.0b5](https://github.com/graphql-python/gql/releases/tag/v3.5.0b5).

PR #429, opened 2023-08-16
Remove invalid arguments mentioned in validation.rst

natekspencer:


natekspencer:
NVM, I was looking at latest, not stable/3.4

Issue #430, opened 2023-09-01
Batching


itolosa:
Hi there,

I've been working with this library for a while, and I'm using it to query an endpoint that is capable of receive multiple queries at once. The API returns responses in the same order as requested.

I've found this feature is implemented by other libraries under the name of Batching. And I've successfully implemented this in Python too. My question is, have you considered this already, are you accepting PRs for this?

I'd be glad to contribute

leszekhanusz:
Something like [transport-level batching](https://www.apollographql.com/blog/apollo-client/performance/query-batching/#transport-level-batching)? Yes, that could be useful.

- Do you know a public GraphQL server configured for this that we could use for tests?
- On which transport do you plan to implement this?

You would probably need to create `execute_batch`, `execute_batch_sync`, `execute_batch_async` methods in client.py which have the same arguments as the `execute` methods but taking in argument a list of `documents`, `variable_values` and `operation_names` instead of a single element and returning a list of answers.

Any help is greatly appreciated. Please check the [CONTRIBUTING.md](https://github.com/graphql-python/gql/blob/master/CONTRIBUTING.md) file. The PR needs to have 100% test code coverage to be accepted.

itolosa:
> Do you know a public GraphQL server configured for this that we could use for tests?

I have one from a crypto exchange. We just need to generate a token but's pretty straightforward, I've managed to obtain one using a temporary email. 

> On which transport do you plan to implement this?

I'd create another transport that derives from the http / requests. In my case BatchingTransport

Actually, I have a working prototype of this on another public repo with 100% coverage, using the github ci tool. I've used github secrets to store the api token.

I've used threads and futures to implement this, but we could do better than that. As a prototype though it's ok.

You have to use the execution result instead of dict to keep an open communication within the transport processor.

Check out this:

https://github.com/itolosa/orionx-api-client/blob/master/orionx_api_client/transports/batch.py

The client should care to not wait for the data from the execution result, So part of the validation should be split to outside of it, as a helper function. In my case called `as_completed`

itolosa:
I'm sorry, I've realized I haven't explained enough what I did.

[...]

PR #431, opened 2023-09-04
Add sync batching to requests sync transport

itolosa:
### Feature
Enable batching.

It adds `execute_batch` method for `Transport` and `Client` in order to allow users to send multiple requests at once.

Server must allow this type of requests.

Issue: #430 

### Resources:
- https://github.com/graphql/graphql-over-http/blob/main/rfcs/Batching.md
- https://spec.graphql.org/October2021


itolosa:
No clue why some tests are failing because of some missing `aiohttp`. It just happens on the CI

leszekhanusz:
> No clue why some tests are failing because of some missing `aiohttp`. It just happens on the CI

In the CI, we are testing that some tests are supposed to work only with a single transport dependency installed.

If you run `pytest tests --requests-only`, it should work even if aiohttp is not installed (after running `pip uninstall aiohttp`)

It is possible to mark the tests to specify which test can be run in which condition. But I just see that you found out about it just now.

docs/code_examples/console_async.py
```diff
@@ -2,7 +2,6 @@
 import logging
 
 from aioconsole import ainput
-
```
leszekhanusz, line 5: Why are all those blank lines removed?
If it is because of isort, flake8 or black, please make sure that you have the correct dev dependencies by running `pip install -e .[dev]`. The dev dependencies are pinned on a specific version to avoid problems like this.

gql/client.py
```diff
@@ -27,6 +28,8 @@
     validate,
 )
 
+from gql.transport.data_structures.graphql_request import GraphQLRequest
```
leszekhanusz, line 31: I would prefer to have a flat folder structure. Please move the `graphql_request.py` file in the `gql` folder and remove the `data_structures` folder. Also `GraphQLRequest` could be imported directly from gql.

gql/transport/data_structures/graphql_request.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any, Dict, Optional
+
+from attr import dataclass
```
leszekhanusz, line 3: Why don't you import `dataclass` from `dataclasses` instead of `attr`?
At the current version we should support Python from version 3.7

gql/transport/httpx.py
```diff
@@ -17,6 +17,8 @@
 import httpx
 from graphql import DocumentNode, ExecutionResult, print_ast
 
+from gql.transport.data_structures import GraphQLRequest
```
leszekhanusz, line 20: Please remove any modification from the httpx.py file. I'll explain below.

gql/transport/requests.py
```diff
@@ -275,6 +274,152 @@ def raise_response_error(resp: requests.Response, reason: str):
             extensions=result.get("extensions"),
         )
 
+    def execute_batch(  # type: ignore
+        self,
+        reqs: List[GraphQLRequest],
+        timeout: Optional[int] = None,
+        extra_args: Dict[str, Any] = None,
```
leszekhanusz, line 281: should be `Optional[Dict[str, Any]]`. I noticed that the same mistake was already present in the execute method and that could be corrected also.

gql/transport/requests.py
```diff
@@ -275,6 +274,152 @@ def raise_response_error(resp: requests.Response, reason: str):
...
+        """
+
+        if not self.session:
+            raise TransportClosed("Transport is not connected")
+
+        # Using the created session to perform requests
+        response = self.session.request(
+            self.method,
+            self.url,
+            **self._build_batch_post_args(reqs, timeout, extra_args),  # type: ignore
```
leszekhanusz, line 305: Please remove the `# type: ignore` here

gql/transport/transport.py
```diff
@@ -17,6 +20,24 @@ def execute(self, document: DocumentNode, *args, **kwargs) -> ExecutionResult:
             "Any Transport subclass must implement execute method"
         )  # pragma: no cover
 
+    @abc.abstractmethod
```
leszekhanusz, line 23: The `execute_batch` method cannot be an `abstractmethod` here.
Doing so would break all the existing custom transports made by users which does not contain the method.

gql/transport/transport.py
```diff
@@ -17,6 +20,24 @@ def execute(self, document: DocumentNode, *args, **kwargs) -> ExecutionResult:
...
+    ) -> List[ExecutionResult]:
+        """Execute multiple GraphQL requests in batch.
+
+        Execute the provided requests for either a remote or local GraphQL Schema.
+
+        :param reqs: GraphQL requests as an iterable of GraphQLRequest objects.
+        :return: iterable of ExecutionResult
+        """
+        raise NotImplementedError(
+            "Any Transport subclass must implement execute_batch method"
```
leszekhanusz, line 38: The text could be modified to something like: "This Transport has not implemented the execute_batch method"

tests/custom_scalars/test_money.py
```diff
@@ -736,6 +758,32 @@ def test_code():
...
+
+    def test_code():
+        with Client(schema=schema, transport=transport, parse_results=True) as session:
+
+            query = gql("query myquery($money: Money) {toEuros(money: $money)}")
+
+            variable_values = {"money": Money(10, "DM")}
+
+            results = session.execute_batch(
+                [GraphQLRequest(document=query, variable_values=variable_values)],
```
leszekhanusz, line 777: Maybe we could execute two requests (the same one twice) in the batch here to show that batch is working with multiple queries.

gql/client.py
```diff
@@ -880,6 +937,112 @@ def execute(
...
+                )
+
+            assert (
+                result.data is not None
+            ), "Transport returned an ExecutionResult without data or errors"
+
+        if get_execution_result:
+            return results
+
+        return [result.data for result in results]  # type: ignore
```
leszekhanusz, line 1044: Please remove the `# type: ignore` here.
I realize that the problem is that mypy is not smart enough to realize that the answer is of type `List[Dict[str, Any]]` instead of `List[Optional[Dict[str, Any]]]` even with the asserts above.
This should be solved by using the `cast` method to cast the answer to `List[Dict[str, Any]]`

tests/test_httpx.py
```diff
@@ -33,6 +34,21 @@
 )
 
 
+@pytest.mark.aiohttp
```
leszekhanusz, line 37: I believe those two marks are not necessary?

leszekhanusz (changes requested):
This look good already!
Also nice to see that the countries.trevorblades.com site we use for our tests supports batch requests.
I've noted a few requested changes.
If you could also please check again all the docstrings so that the parameters correspond to the new ones and the text is modified to say that it's a batch operation.
Thanks!

gql/client.py
itolosa: whoops, my mistake you're completely right!

gql/transport/data_structures/graphql_request.py
itolosa: agree, vscode autocompleted that for me 😷

gql/transport/httpx.py
itolosa: Got it

gql/transport/requests.py
itolosa: Of course! I've just copied that from the old code

gql/transport/transport.py
itolosa: agree, better to leave the check in runtime to avoid to always implement this

gql/transport/transport.py
itolosa: yes, thank you, I need to modify all the docstrings to match the new code

tests/custom_scalars/test_money.py
itolosa: ok, seems fair

gql/client.py
itolosa: Agree, but anyways I don't really like the type system of python. it's horrible, I tried to not worry too much about it because it simply can't handle any simple case. But no problem about the casting.

tests/test_httpx.py
itolosa: yes, I've realized it now, since the method doesn't use async anymore

docs/code_examples/console_async.py
itolosa: I've done exactly that, but I don't know what's wrong with my local setup. I'm having other errors also because of some failing tests from the `test_cli.py` file. Anyways, better solution for now is to revert those changes

leszekhanusz: Are you sure you used the `-e` flag with pip? I know that if I don't use it, then the code coverage is skipping some lines (510-539) from `cli.py` for some reason...

itolosa: Ok I'll try with bash next time, I think zsh is incorrectly parsing the command. It should be it

gql/client.py
leszekhanusz: I know, but it helps sometimes. I prefer to avoid using ignore as much as possible if there is a clean way to do it, it avoids the mental load when you read the code months later and you don't know if it's legit to ignore the type here.

docs/code_examples/console_async.py
leszekhanusz: Yes, I believe with zsh you need to use quotes: `pip install -e ".[dev]"`

itolosa (commented):
Pushed some fixes

gql/client.py
```diff
@@ -27,6 +28,8 @@
     validate,
 )
 
+from gql.graphql_request import GraphQLRequest
```
leszekhanusz, line 31: Please use a relative import (`from .graphql_request`) here.

gql/transport/requests.py
```diff
@@ -275,6 +274,152 @@ def raise_response_error(resp: requests.Response, reason: str):
             extensions=result.get("extensions"),
         )
 
+    def execute_batch(  # type: ignore
+        self,
+        reqs: List[GraphQLRequest],
+        timeout: Optional[int] = None,
+        extra_args: Optional[Dict[str, Any]] = None,
+    ) -> List[ExecutionResult]:
+        """Execute GraphQL query.
```
leszekhanusz, line 283: Execute multiple GraphQL requests in a batch

gql/transport/requests.py
```diff
@@ -275,6 +274,152 @@ def raise_response_error(resp: requests.Response, reason: str):
...
+
+        Execute the provided requests against the configured remote server. This
+        uses the requests library to perform a HTTP POST request to the remote server.
+
+        :param reqs: GraphQL requests as a list of GraphQLRequest objects.
+        :param timeout: Specifies a default timeout for requests (Default: None).
+        :param extra_args: additional arguments to send to the requests post method
+        :param upload_files: Set to True if you want to put files in the variable values
```
leszekhanusz, line 291: upload_files param is removed

gql/transport/transport.py
```diff
@@ -1,7 +1,10 @@
 import abc
+from typing import List
 
 from graphql import DocumentNode, ExecutionResult
 
+from gql.graphql_request import GraphQLRequest
```
leszekhanusz, line 6: Please use a relative import here.

tests/test_graphql_request.py
```diff
@@ -0,0 +1,203 @@
...
+    GraphQLList,
+    GraphQLNonNull,
+    GraphQLObjectType,
+    GraphQLScalarType,
+    GraphQLSchema,
+)
+from graphql.utilities import value_from_ast_untyped
+
+from gql import gql
+from gql.graphql_request import GraphQLRequest
```
leszekhanusz, line 22: You could import GraphQLRequest from gql directly on the line above.

leszekhanusz (changes requested):
a few nitpicking changes.

leszekhanusz (approved):
LGTM

itolosa:
Nice work! 😄

leszekhanusz:
Thanks for your help!

PR #436, opened 2023-09-09
Sync auto batching requests

leszekhanusz:
Implementation of an automatic batching of requests for the sync transports.

See issue #430

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -1107,6 +1107,7 @@ class SyncClientSession:
             except Exception as exc:
                 for future in futures:
                     future.set_exception(exc)
+                continue
 
             # Fill in the future results
             for result, future in zip(results, futures):
--- a/gql/client.py
+++ b/gql/client.py
@@ -1046,6 +1046,7 @@ class SyncClientSession:
         *,
         serialize_variables: Optional[bool] = None,
         parse_result: Optional[bool] = None,
+        validate_document: Optional[bool] = True,
         **kwargs,
     ) -> List[ExecutionResult]:
         """Execute multiple GraphQL requests in a batch, using
@@ -1057,13 +1058,16 @@ class SyncClientSession:
             By default use the serialize_variables argument of the client.
         :param parse_result: Whether gql will unserialize the result.
             By default use the parse_results argument of the client.
+        :param validate_document: Whether we still need to validate the document.
 
         The extra arguments are passed to the transport execute method."""
 
         # Validate document
         if self.client.schema:
-            for req in requests:
-                self.client.validate(req.document)
+
+            if validate_document:
+                for req in requests:
+                    self.client.validate(req.document)
 
             # Parse variable values for custom scalars if requested
             if serialize_variables or (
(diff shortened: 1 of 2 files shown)


gql/client.py
```diff
@@ -1038,6 +1184,105 @@ def execute_batch(
...
+            if self.batch_queue.qsize() < self.client.batch_max - 1:
+                time.sleep(self.client.batch_interval)
+
+            # Then get the requests which had been made during that wait interval
+            for _ in range(self.client.batch_max - 1):
+                if self.batch_queue.empty():
+                    break
+                request_and_future = self.batch_queue.get()
+                if request_and_future is None:
+                    stop_loop = True
```
itolosa, line 1213: Another possible option here is to wrap the while loop, with try - except, and raise an exception in here. If requests_and_futures is not empty, set_exception for each future. I'll leave the decision to you if you want to do it or not. I consider this way is also fine, if the user wants to perform the last requests after the termination.

tests/test_requests_batch.py
```diff
@@ -375,3 +532,134 @@ def test_code():
...
+        for continent_code in continent_codes:
+
+            thread = Thread(
+                target=get_continent_name,
+                args=(
+                    session,
+                    continent_code,
+                ),
+            )
+            thread.start()
```
itolosa, line 597: don't we should join the thread after it's started?

tests/test_requests_batch.py
```diff
@@ -74,6 +89,108 @@ def test_code():
...
+                assert africa["code"] == "AF"
+
+                # Checking response headers are saved in the transport
+                assert hasattr(transport, "response_headers")
+                assert isinstance(transport.response_headers, Mapping)
+                assert transport.response_headers["dummy"] == "test1234"
+
+            for _ in range(2):
+                thread = Thread(target=test_thread)
+                thread.start()
```
itolosa, line 189: should we wait for the thread?

itolosa (changes requested):
Just a tiny change, the rest are question or suggestions. Also, I don't know if we should support the empty list as a valid argument in the requests parameter. If not, we should include an assert in the client

gql/client.py
leszekhanusz: It is necessary to wait for all the requests to be made or else the promise could never resolve.

gql/client.py
```diff
@@ -146,6 +156,12 @@ def __init__(
 
         self.serialize_variables = serialize_variables
         self.parse_results = parse_results
+        self.batch_interval = batch_interval
+        self.batch_max = batch_max
```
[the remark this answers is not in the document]
leszekhanusz: I'm going to leave it as it for consistency is as we currently don't validate the other arguments either.

tests/test_requests_batch.py
leszekhanusz: I don't think it's really necessary but it does not hurt, so I changed it.

tests/test_requests_batch.py
leszekhanusz: I don't think it's really necessary but it does not hurt, so I changed it.

PR #449, opened 2023-11-19
Adding node_tree method in utilities to debug and compare DocumentNode instances

leszekhanusz:
The `node_tree` method allows to see the hierarchy of a DocumentNode object.

Now the DSL code will not set variable definitions for Fragments by default to be coherent with the default graphql-core parsing in the `gql` method.


leszekhanusz:
@cito I don't know if you have something similar for graphql-core but I think the `node_tree` method could be useful to you too to debug deep DocumentNode objects.

PR #453, opened 2023-12-13
Decrease log level to debug on message/answer

ztamas83:
Info logs are too verbose, messages are logged every second in the pyTibber package and probably others as well that are using this lib.


leszekhanusz:
Sorry, I can't accept this PR as this is a breaking change.
I might do this for the next major version (currently not planned).

See issue #178 

You can decrease the log level for gql by following the [instructions here](https://gql.readthedocs.io/en/stable/advanced/logging.html).

You could instead make a PR in pyTibber to reduce gql logs there.

Issue #457, opened 2024-01-02
Tests fail out of the box.


MrSampson:
I wanted to make a small change, so I followed the contributing guidelines and:

`make tests` passes 100% but `make all_tests` fails on 18 tests.

Is this to be expected? 

BTW, what I wanted to change is related to the aiohttp 3.9 restriction to python3.12 I'd like to change that to 3.11 since I'm unable to upgrade to 3.12 at the moment. aiohttp 3.9 is necessary due to the two CVEs that came out a month ago.

https://nvd.nist.gov/vuln/detail/CVE-2023-49081
https://nvd.nist.gov/vuln/detail/CVE-2023-49082

Thanks,
Oliver

leszekhanusz:
> `make tests` passes 100% but `make all_tests` fails on 18 tests.
> Is this to be expected?

Yes, `make all_tests` run some tests on a backend out of our control. That's why they are not part of the default tests as they could be flaky and the backend could change at any time.

I made the PR #459 to fix them for now.

See your PR for your other issue.

MrSampson:
Ah, cool. Thanks!

PR #458, opened 2024-01-02
Change aiohttp3.9 python3.12 requirement

MrSampson:
Due to two CVEs, downstream products that may not have upgraded to python3.12 need aiohttp to be at version 3.9.

The update changes the requirement for aiohttp==3.9 to include python>=3.11

See also:
https://nvd.nist.gov/vuln/detail/CVE-2023-49081
https://nvd.nist.gov/vuln/detail/CVE-2023-49082

I've verified the tests and no more tests are failing than before the changes. See https://github.com/graphql-python/gql/issues/457

Thanks,
Oliver

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,8 @@ dev_requires = [
 ] + tests_requires
 
 install_aiohttp_requires = [
-    "aiohttp>=3.8.0,<4;python_version<='3.11'",
-    "aiohttp>=3.9.0b0,<4;python_version>'3.11'",
+    "aiohttp>=3.8.0,<4;python_version<'3.11'",
+    "aiohttp>=3.9.0b0,<4;python_version>='3.11'",
 ]
 
 install_requests_requires = [


leszekhanusz:
I don't really understand what you want to do.

This PR is not necessary to install `gql` with `aiohttp` version 3.9 on Python 3.11.
You can do this with: `pip install --pre gql[aiohttp] aiohttp==3.9.0b0` among other methods.

MrSampson:
I'm using poetry to manage the dependencies, and when I set aiohttp"^3.9.0" it cannot resolve the dependencies, because aiohttp is only available for python3.12:

```
poetry show --tree gql
gql 3.4.1 GraphQL client for Python
├── aiohttp >=3.7.1,<3.9.0
│   ├── aiosignal >=1.1.2 
│   │   └── frozenlist >=1.1.0 
│   ├── async-timeout >=4.0.0a3,<5.0 
│   ├── attrs >=17.3.0 
│   ├── charset-normalizer >=2.0,<4.0 
│   ├── frozenlist >=1.1.1 (circular dependency aborted here)
│   ├── multidict >=4.5,<7.0 
│   └── yarl >=1.0,<2.0 
│       ├── idna >=2.0 
│       └── multidict >=4.0 (circular dependency aborted here)
├── backoff >=1.11.1,<3.0
├── botocore >=1.21,<2
│   ├── jmespath >=0.7.1,<2.0.0 
│   ├── python-dateutil >=2.1,<3.0.0 
│   │   └── six >=1.5 
│   └── urllib3 >=1.25.4,<2.1 
├── graphql-core >=3.2,<3.3
├── requests >=2.26,<3
│   ├── certifi >=2017.4.17 
│   ├── charset-normalizer >=2,<4 
│   ├── idna >=2.5,<4 
│   └── urllib3 >=1.21.1,<3 
├── requests-toolbelt >=0.9.1,<1
│   └── requests >=2.0.1,<3.0.0 
│       ├── certifi >=2017.4.17 
│       ├── charset-normalizer >=2,<4 
│       ├── idna >=2.5,<4 
│       └── urllib3 >=1.21.1,<3 
├── urllib3 >=1.26,<2
├── websockets >=10,<11
└── yarl >=1.6,<2.0
    ├── idna >=2.0 
    └── multidict >=4.0 
```
Why is there a python3.12 requirement for the aiohttp dependency?

leszekhanusz:
You've got it backwards. Those lines in the `setup.py` file means:

- If you're using Python 3.11 or earlier, aiohttp should be at least 3.8 and below 4
- If you're using Python 3.12, aiohttp should be at least 3.9.0b0 and below 4

It does NOT mean that you have to use Python 3.12 if you are using aiohttp 3.9. In all versions of Python you can use aiohttp version 3.9

That line was added in the `setup.py` file because when Python 3.12 support was added, the 3.8 version of aiohttp was not working on Python 3.12 and a version 3.9.x of aiohttp is required if you use Python 3.12

The reason Poetry is restricting you to an earlier version of aiohttp is because you are using the stable version of gql, version 3.4.1
On this version, the setup.py [looks like this](https://github.com/graphql-python/gql/blob/3434b3f16dee5f37523b6c1aa86dc71e10f85310/setup.py#L41), and aiohttp is restricted to a 3.8.x version of aiohttp for all versions of Python.


To solve your problem, you have to use the pre-release version of gql.

To install the pre-release version of gql using poetry you can use:
`poetry add gql --allow-prereleases`

That command will modify your pyproject.toml file to have the following line:
`gql = {version = "^3.5.0b9", allow-prereleases = true}`

MrSampson:
Oh, okay, thank you. I'll look into that. I was indeed looking in the wrong branch.

However, would it make sense to release a version (3.4.2) that allowed aiohttp < 4 for python < 3.12, specifically to allow the the CVE fixes to come downstream automatically? I'd be happy to push that PR to you.

MrSampson:
Thank you!

Issue #451, opened 2023-11-27
gql should have a websockets-client transport


mpegman-scwx:
python-websockets doesn't support proxies and has no intention to in the near future.  We should allow for usage of a websockets-client transport to allow users to ability to run through proxies.

I plan on working on this myself.  Just creating a placeholder issue to work off of, but I am willing to work with anyone else that would be interested in supporting this feature.

leszekhanusz:
Wouldn't it be better to make a new websockets transport using `aiohttp` as proposed by #418 instead of using adding a new `websockets-client` dependency?

Have you looked into [this comment](https://github.com/python-websockets/websockets/issues/475#issuecomment-1268504779) for a way to make a websocket connection through a proxy using `websockets`?

mpegman-scwx:
I was looking through aiohttp docs and everything there uses the `async` keyword.  It would be nice to have a websockets option that supported proxies and didn't need event loop management.

leszekhanusz:
Well, the great thing about gql is that you can use any transport (sync or async) in a sync way if you want. If you look at the [default subscription example](https://gql.readthedocs.io/en/stable/usage/subscriptions.html), you don't see any async code there.

And what about the method in the comment above? Can't it solve your problem?

mpegman-scwx:
It's worth to looking into, but at the moment hasn't been implemented yet.  It would take me some time to familiarize myself with aiohttp (something I haven't used at all and don't know any of its internals).  Looking through the `subscribe()` method it looks like it needs an async transport anyways, so maybe this wouldn't even integrate well.

aaugustin:
>  It would be nice to have a websockets option that supported proxies and didn't need event loop management.

You will have both in version 15.0 in a few weeks.

Issue #418, opened 2023-06-06
aiohttp transport should support websockets


devkral:
aiohttp has out of the box websocket support.

So it should be easy to integrate in the transport.

leszekhanusz:
Sure, why not make a new transport named `AIOHTTPWebsocketsTransport`.
I'll accept a PR if it is well-tested and comes with 100% test coverage.

PR #478, opened 2024-05-09
New transport AIOHTTPWebsocketsTransport

tlowery-scwx:
This represents ongoing work addressing issues #418 and #451, adding AIOHTTP Websockets support with a new AIOHTTPWebsocketsTransport. We already have a working base model, and are working on full parity and test coverage, and wanted to make this work available for review or comment. 

This transport allows to run queries including subscriptions on websockets endpoints without using the `websockets` dependency (only the `aiohttp` dependency).

Note: a lot of code had to be duplicated from the `WebsocketsTransport` class. A refactor will be done to reduce the duplicated code in a subsequent PR.


leszekhanusz:
Thanks for your PR.
To fix the lint issue, you can use `make check` (See [CONTRIBUTING.md](https://github.com/graphql-python/gql/blob/master/CONTRIBUTING.md#coding-guidelines))

tlowery-scwx:
Hi @leszekhanusz, I'm wondering if you could give us some guidance on the final stage of this PR. We've been pursuing 100% test coverage, and we're down to a single edge case where a `TransportClosed` exception would be raised [here](https://github.com/tlowery-scwx/gql/blob/bae308f07657ae32c077d86698e3da526f7e3264/gql/transport/aiohttp_websockets.py#L534) and bubble up [here](https://github.com/tlowery-scwx/gql/blob/bae308f07657ae32c077d86698e3da526f7e3264/gql/transport/aiohttp_websockets.py#L653). In our case, in order to implement a Transport using AIOHTTP rather than Websockets, we had to duplicate, rather than inherit from, WebsocketsTransportBase. We've been having trouble replicating the circumstances under which the mentioned edge cases would occur, and on further investigation, we found that the corresponding lines in WebsocketsTransportBase are only covered by the tests for the PheonixChannelWebsocketsTransport -- and that [their test](https://github.com/tlowery-scwx/gql/blob/bae308f07657ae32c077d86698e3da526f7e3264/tests/test_phoenix_channel_exceptions.py#L194) which covers this functionality hangs indefinitely when a WebsocketsTransport is substituted for a PheonixChannelWebsocketsTransport. Since PRs require 100% test coverage, we're at something of an impasse -- can you recommend a course of action for addressing this edge case which is not tested by the base class?

leszekhanusz:
I'll  take a look soon.

leszekhanusz:
Hi @tlowery-scwx. Thanks for your work so far.

I took the liberty to take over your branch to try to fix the remaining problems, you can check the commits above.

I had to copy yet again some code from the previous `websockets_base.py` file (`ListenerQueue`...) so that it would be possible to run the new transport without needing the `websockets` dependency. I'll proceed in two steps, first put everything in the `aiohttp_websockets.py` file and make it work in this PR, then when this PR has been merged, I'll refactor everything to follow DRY in another PR.

The `aiohttp_websocket` pytest mark has been removed. There was a bit of confusion here, the goal of those pytest marks are there to select tests based on the installed dependencies, so only `aiohttp` and `websockets` marks make sense. Something annoying now is that we still need the websockets dependency to test the new transport as we still use websockets for the server part in the tests, so both marks are set for the tests of the new transport. I'll try to implement a new websocket server using the aiohttp dependency so that we could be sure that all tests are passing without the previous websockets dependency.

One of the main problem was that the aiohttp session was not closed properly. For that it is a bit tricky because of still remaining aiohttp issues which should be fixed in aiohttp 4.0 I suppose. Fortunately we are already doing that for our aiohttp transport so I copied the code from there.

During the testing, it appeared that the aiohttp session was still not closed properly for some tests, and it revealed that the problem was already present for the previous websockets transport for some conditions. I fixed that in the PRs #486 and #488, which have been merged in this branch and made the required changes in this branch.

Regarding the edge case with the `TransportClosed` exception in the `_receive` method, to fix the code coverage, at first I simply replaced that check with an assertion, as there is not point to have code for which we don't know how to trigger it. I see now that you have made a test which is skipped where you set the `transport.websocket` to `None` and discovered other problems here (hanging indefinitely). I'm now thinking reverting that change and reinforcing the `_close_coro` method so it is more resilient. And instead of checking if `self.websocket` is `None` in the `_fail` method (to see if the transport has already been closed), we can instead check the status of the `self._wait_closed` event instead. I'll implement that in a following commit. We should also add a configurable timeout for waiting for the close in case of unforeseen problems there.

If you don't mind I'll continue to work on this branch for a few days before asking for your review so please don't add new commits for the time being.

leszekhanusz (approved):

leszekhanusz:
@tlowery-scwx It should be ok now. What do you think?

tlowery-scwx:
@leszekhanusz, thank you so much for taking this on, putting so much work into it, and getting it over the finish line! May we have a day to test this transport with our use case and get back to you by EOD tomorrow (USA Pacific time) before you merge and close this out?

tlowery-scwx:
Hi  @leszekhanusz -- sorry for taking a bit longer than planned, but we've done quite a bit of testing, and everything looks great from our end. Thank you again for all of your work and help getting this done!

PR #483, opened 2024-06-12
Update annotation for client.execute_batch, get_execution_result argu…

rahul-gj:
…ment.

client.execute_batch  was showing "No overload for provided arguments" in Pylance for vscode. Had to provide argument for get_execution_result to suppress error. Update overloads to have default values to overcome this.


leszekhanusz:
@rahul-gj Sorry for the delay. I modified the `get_execution_result` argument typing so that it is identical to the other methods in the file. I believe this could solve your problem.

I don't use vscode or pylance so if you could verify that would be nice.

Thanks for the report!

PR #507, opened 2024-10-28
Revert codecov action to v1

leszekhanusz:


leszekhanusz:
Now solved with v4 with correct token

PR #511, opened 2024-11-16
types-mock was only usefull when using old standalone "mock" module

a-detiste:


leszekhanusz:
Thanks. I'll investigate the coverage problems which have nothing to do with this PR before merging this one.

leszekhanusz (approved):

PR #513, opened 2024-12-11
Wire httpx transport in gql-cli

mss:
I thought that I had some problem with the `aiohttp` backend and tried to use `httpx` instead just to discover that it wasn't available.

My problem was somewhere else but this change makes the `httpx` backend available for the `gql-cli` command.


mss:
Ok, now I know why that other line was formatted so weirdly, to make flake8 happy. I will revert that line.

mss:
I also added a test so codecov is hopefully happy again.

leszekhanusz (approved):
Thanks!

PR #514, opened 2024-12-11
Add minimal pyproject.toml

mss:
My `setuptools` knowledge is a bit rusty, especially when it comes to how to use the entry points while developing.

This change adds a minimal `pyproject.toml` which makes a call like `uv run gql-cli ...` available.


leszekhanusz (approved):

leszekhanusz:
Thanks, should have done this years ago.

PR #528, opened 2025-01-30
Use httpx with gql-cli if aiohttp is not available on auto transport

leszekhanusz:
Using the httpx transport by default with gql-cli if aiohttp is not available if no transport is specified.

Related: #527

### DIFF
--- a/gql/cli.py
+++ b/gql/cli.py
@@ -297,7 +297,23 @@ def autodetect_transport(url: URL) -> str:
 
     else:
         assert url.scheme in ["http", "https"]
-        transport_name = "aiohttp"
+
+        try:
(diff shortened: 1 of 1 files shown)


traylenator:
```
echo 'query { continent(code:"AF") { name } }' | gql-cli https://countries.trevorblades.com
Error: 

No suitable dependencies has been found for an http(s) backend (aiohttp or httpx).

Please check the install documentation at:
https://gql.readthedocs.io/en/stable/intro.html#installation
```

Thanks.

leszekhanusz:
@traylenator What is the problem? Did you check the installation instructions?

traylenator:
> @traylenator What is the problem? Did you check the installation instructions?

Nothing is wrong - was just saying thanks . All good.

Issue #529, opened 2025-02-17
AIOHTTPTransport ssl=False by default


mariuskava:
**Common problems**
- TLS should be enabled by default. AIOHTTPTransport uses ssl=False by default
- I was expecting library to fail when using self signed certificate on the server.

**To Reproduce**
- start a server with a self signed certificate
- try connect
- it will succeed instead of failing

**Expected behavior**
Expect connection to fail when using not trusted certificates

**System info (please complete the following information):**
 - Any

leszekhanusz:
That's pretty bad... Thanks for your report.

The annoying thing is that there are probably some people which are depending on this and changing the default will be a breaking change for them.

A major version bump will be needed to fix this.

mariuskava:
yes, there must be a breaking change.. but a patch release could contain a warning:
- default should be changed to ssl=None
- if ssl==None: log warning and set ssl=False

leszekhanusz:
> yes, there must be a breaking change.. but a patch release could contain a warning:

Good idea!

leszekhanusz:
Note that for the warning, the default value cannot be `None`.

That's because there was a change in aiohttp from version 3.10

Before aiohttp 3.10, aiohttp was using `None` as the default value, and would verify the certification in that case.
From aiohttp 3.10, it switched to use `True` as the default value.

So, someone could have used `ssl=None` previously with `AIOHTTPTransport`. He would have a typing error with mypy but it would work and gql would verify the certificate. If we set ssl to False if we receive None, then gql will not verify the cert for him in that case.

We could use another default value, like the "warning" string?

mariuskava:
good point. Default string value sounds good enough or it could be a predefined sentinel = object()
https://python-patterns.guide/python/sentinel-object/#sentinel-objects

PR #530, opened 2025-02-17
AIOHTTPTransport default ssl cert validation add warning

leszekhanusz:
Issue #529 reported that the `AIOHTTPTransport` did not validate ssl certificates by default.

This PR add tests for all transports to check the four possible cases using a self certificate:

 - By default, and if the ssl parameter is set to True, the validation should be made and fail.
 - if the ssl parameter is set to the self-certificate ssl context, the validation should be made and succeed.
 - if the ssl parameter is set to False, the validation should not be made and the request should succeed.

A new major version will be made to change this default behavior. In the meantime This PR will add a warning for `AIOHTTPTransport` for the default case to warn users.

Note that the tests also revealed that it was not possible to disable the certificate validation on the `WebsocketsTransport`. This should also be changed in the next major version.


gql/transport/aiohttp.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         headers: Optional[LooseHeaders] = None,
         cookies: Optional[LooseCookies] = None,
         auth: Optional[Union[BasicAuth, "AppSyncAuthentication"]] = None,
-        ssl: Union[SSLContext, bool, Fingerprint] = False,
+        ssl: Union[SSLContext, bool, Fingerprint] = True,
```
mariuskava, line 49: does it still work well with http:// urls like http://localhost:8888/graphql?
(cannot test right now on my phone..)

leszekhanusz: Yes, most of the tests are not using `https` and are all passing with this.

PR #543, opened 2025-03-12
Upgrade lastest websockets and Exceptions overhaul

leszekhanusz:
- Using new asyncio implementation of the `websockets` dependency.
- Bump `websockets` to >=14.2
- Reduced some timers to make tests faster.
- Now the reconnecting session will reconnect as soon as it detects that the connection failed.
- Better log messages
- using `TransportConnectionFailed` instead of `TransportClosed` for exceptions which were due to the connection failing or closing. Using `TransportClosed` only for a closing of the transport initiated by the user.

### DIFF
--- a/gql/client.py
+++ b/gql/client.py
@@ -35,7 +35,7 @@ from graphql import (
 
 from .graphql_request import GraphQLRequest
 from .transport.async_transport import AsyncTransport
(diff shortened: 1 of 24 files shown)


leszekhanusz (commented):
:+1:

PR #573, opened 2025-11-12
chore: Replace archived backoff with tenacity

Enimalojd:
The backoff library is no longer maintained. This commit replaces it with tenacity, which provides equivalent functionality for retry logic.

Changes:
- Replace backoff dependency with tenacity>=9.1.2 in setup.py
- Update retry decorators in gql/client.py to use tenacity API
- Update all documentation examples to use tenacity instead of backoff
- Update code examples to use tenacity retry decorators

The functionality remains the same:
- retry_connect: exponential backoff with max 60 seconds (infinite retries)
- retry_execute: exponential backoff, 5 attempts, excluding TransportQueryError


leszekhanusz:
Thanks!

leszekhanusz (approved):

PR #591, opened 2026-04-23
Add extensions support to GraphQLRequest

### LINKED ISSUE #590: Support extensions on requests
Currently, GraphQLRequest.payload only includes query, variables, and operationName

Let's add support for extensions? (e.g. to support persisted operation ids)

magicmark:
## Summary

Adds an optional `extensions` parameter to `GraphQLRequest`, as defined in the [GraphQL over HTTP spec](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md). When set, it is included as a top-level key in the request payload alongside `query`, `variables`, and `operationName`.

No transport changes needed — all transports already serialize `request.payload`.

```python
request = GraphQLRequest(
    "query { viewer { name } }",
    extensions={"persistedQuery": {"version": 1, "sha256Hash": "abc123"}},
)
result = await session.execute(request)
```

Closes #590

## Test plan

- Unit tests for extensions in payload, copy, override, and serialize_variable_values
- Integration tests for execute, subscribe, and batch across aiohttp, httpx, requests, and websockets

### DIFF
--- a/docs/usage/extensions.rst
+++ b/docs/usage/extensions.rst
@@ -12,7 +12,7 @@ top-level key in the request payload alongside :code:`query`, :code:`variables`,
 and :code:`operationName`.
 
 You can use this to pass protocol extensions such as
-`persisted queries <https://www.apollographql.com/docs/apollo-server/performance/apq/>`_:
+`trusted documents <https://graphql.org/learn/security/#trusted-documents>`_:
 
 .. code-block:: python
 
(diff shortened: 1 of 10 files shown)


magicmark:
@leszekhanusz what do you think about this approach vs a new argument to `session.execute` (and friends), or even `extra_args`?

(extra_args seems to be something intended to be passed directly to the transport though, so that's probably not a good fit)

leszekhanusz:
Looks good to me! Much better than having to add an extensions attribute to ALL the execute and subscribe methods.

leszekhanusz (approved):

leszekhanusz:
Thanks!

PR #598, opened 2026-06-30
Skip multipart parts with empty Content-Type instead of raising

magicmark:
Multipart HTTP streams may contain empty parts (no headers, no body):

> NO header fields are actually required in body parts. A body part that starts with a blank line, therefore, is allowed
~ _https://www.rfc-editor.org/info/rfc2046/#section-5.1_

(In our case, we see this (I think) due to being passed through multiple proxies which unpack and pack parts back together)

Anyway our current handling is too strict — it required every part to have `Content-Type: application/json` instead of just skipping empty ones.

### Before

When this happens, applications see traceback like this:

```
gql.transport.exceptions.TransportProtocolError: Unexpected part content-type: . Expected 'application/json'.
Unexpected part content-type: . Expected 'application/json'.
```

### After

✨ empty parts are ignored; no errors

### DIFF
--- a/gql/transport/aiohttp.py
+++ b/gql/transport/aiohttp.py
@@ -593,8 +593,7 @@ class AIOHTTPTransport(AsyncTransport):
             errors = data.get("errors")
             if errors:
                 error_messages = [
-                    error.get("message", "Unknown transport error")
-                    for error in errors
+                    error.get("message", "Unknown transport error") for error in errors
                 ]
 
                 for message in error_messages:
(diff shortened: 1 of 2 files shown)


leszekhanusz:
According to the appolo graphql multipart documentation, an empty json object should be sent to specify a heartbeat: https://www.apollographql.com/docs/graphos/routing/operations/subscriptions/multipart-protocol#heartbeats

leszekhanusz:
Please modify the code to [only skip a part without content-type if it is empty](https://chatgpt.com/share/6a43b179-9194-83eb-9d31-147e12a853d8).

leszekhanusz:
And please don't force-push, all the commits will be squashed anyway when the PR is merged.

leszekhanusz (approved):

